PR #304, opened 2026-07-04
Recover TCO-elided frames in uncaught-error stack traces

yogthos:
## What

Uncaught errors now print a useful `trace:` backtrace even on the runtime eval path, and there's an opt-in tail-frame history (`JOLT_TRACE=1`) that recovers the frames tail-call optimization erases.

### Always-on (no cost)
The eval path registers no source map, so `jolt-backtrace-string` used to drop every walkable frame and print nothing. It now keeps any named, non-plumbing continuation frame (bare name when unmapped), so you see the surviving non-tail spine:

```
Unhandled exception: class clojure.lang.Keyword cannot be cast to class java.lang.Number
  at .../repro/core.clj:5:1
  trace:
    middle
    outer
    -main
```

### JOLT_TRACE — tail-frame history (ring of rings)
TCO erases tail-called frames, so the immediate error site is often missing. `JOLT_TRACE=1` records each compiled fn on entry into a bounded **ring-of-rings** (MIT Scheme's "history" shape): the outer ring holds one rib per non-tail subproblem, each rib a small inner ring of the tail-calls at that level. A tight tail loop churns one rib instead of flushing the spine, so the non-tail caller context survives and space stays bounded.

```
;; a<->b tail loop under a non-tail caller (driver/top):
  trace:
    a
    b
    a
    b        <- loop confined to one bounded rib
    driver   <- non-tail caller preserved
    top
```

## How

- Emitter marks a tail call with `(jolt-trace-mark! #t)`; the runtime routes the callee into the current rib vs a fresh one. A `*tail?*` dynamic var tracks tail position (cleared by default, passed through `if`/`do`/`let`/`loop`/fn-body).
- Gated on `trace-frames?`: `compile-eval` turns it on for `JOLT_TRACE`; `emit-image` / `jolt build` force it off. Non-trace emitted output is **byte-identical** (prelude unchanged, seed re-minted to a fixpoint), and a built binary carries no per-call cost.
- History resets per top-level form, so a REPL error isn't padded with earlier frames.

## Tests / validation
- `smoke` 36/36 (adds Tier-0 and `JOLT_TRACE` trace assertions, incl. ring-of-rings caller-preservation and per-form reset)
- `selfhost` fixpoint holds; `remint` converges in 2 passes; prelude byte-unchanged
- `corpus` 3392/3401, **0 crashes / 0 new divergences**; `unit`, `values`, `buildsmoke`, and the numeric/inline/directlink/infer/devirt gates all green

## Known limitation
Tail-frame labeling is best-effort: a tail call routed through `jolt-invoke` to a target with no entry prologue (a core/native fn, an anonymous fn in a var) doesn't consume the mark, so a following frame can be mis-grouped — cosmetic only, never a wrong result. Documented in `rt.ss`.


yogthos:
Added: **tracing on by default in REPL-driven development** (a `repl` or nREPL session enables it; the REPL/nREPL error paths print the tail-frame backtrace via a new history-only `jolt.host/backtrace-string`). `JOLT_TRACE=1` still forces it on for a whole run (incl. `-M:run`'s own load); `JOLT_TRACE=0` forces off. No seed change — `jolt.main`/`jolt.nrepl` are runtime-loaded. Smoke now 38/38.

yogthos:
Post-review updates (all runtime/host or build-driver — no additional seed re-mint):

- **Reviewed the dev-mode commit** (independent pass). Confirmed the backtrace refactor, per-form history reset, nREPL thread-local ring, and `:err` concatenation are correct. Found + fixed a `JOLT_TRACE` footgun: `off`/`FALSE`/`No`/`n` used to *enable* tracing (only exact `0`/`false`/`no` disabled), and the two enable paths disagreed on empty string — now one case-insensitive `jolt-trace-env-off?` predicate.
- **AOT `*ns*` parity**: a built binary ran `-main` with `*ns*` left at the entry ns; `clojure.main` (and interpreted joltc) use `user`, where `(resolve 'alias/sym)` is nil. Reset to `user` in the launcher before `-main`. build-smoke now asserts `ns: user` + `resolve: nil` + `ns-resolve` works, via a separate two-namespace app.

Validation: smoke 38/38, buildsmoke green (incl. ns-parity), selfhost fixpoint, corpus 3392/3401.

PR #307, opened 2026-07-06
Allowed the boot script to check multiple names for the chez executable.

simon-brooke:
This changes only the shell script `bin/joltc`; it checks for an executable named either `chez` or `chezscheme`, in that order. Further possible names could trivially be added.

### DIFF
--- a/bin/joltc
+++ b/bin/joltc
@@ -14,7 +14,28 @@
 # JOLT_PWD.
 root="$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)"
 export JOLT_PWD="${JOLT_PWD:-$PWD}"
+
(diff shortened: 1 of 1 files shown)


yogthos:
ah good call!

simon-brooke:
Uhhmmm... This isn't sufficient, sadly.

Both `Makefile` and `host/chez/selfcheck.sh` also have dependency on Chez Scheme being invoked as `chez`, and there may be other scripts I haven't yet found. I'll work through these, and get them fixed.

This pull request isn't going to do any harm, I don't believe, but it isn't a complete solution.

yogthos:
Ah noted, if you get a chance to dig through a bit a follow up is welcome, and I'll see if I can get to it as well in the near future.

PR #359, opened 2026-07-14
Extract java.time.* into the jolt-lang/time library

yogthos:
Move the java.time.* implementation out of core into the new [jolt-lang/time](https://github.com/jolt-lang/time) library as portable Clojure, and delete `host/chez/java/java-time.ss` (2579 lines).

The library provides the whole `java.time.*` surface — LocalDate/Time/DateTime, Instant, ZonedDateTime, OffsetDateTime, Duration, Period, Year/YearMonth, zones, DateTimeFormatter, the temporal machinery — over the `__register-class-*` seams, and pulls juxt/tick through deps.edn for the idiomatic API. tick's full suite passes (706 assertions) plus the 93 java.time conformance rows migrated out of the core corpus, 1020 assertions total, 0 failures.

Core keeps the `#inst` / `java.util.Date` layer (inst-time.ss) and the irreducible libc time primitives, now factored into `tz-primitives.ss` and exposed as `jolt.host/tz-offset-seconds` / `locale-name`.

New core seams (all runtime .ss, no re-mint):
- `__register-eq!/hash!/str!/pr!/compare!` — a library gives its own host values full value semantics.
- `__register-class!` — those values answer `class`/`type` and dispatch protocols extended to their class (was `:object`; tick extends protocols to java.time classes).
- `java.util.Date` is Comparable; `jolt.host/set-instant-ctor!` unifies `Date.toInstant` with the library's Instant.

Surfaced two general jolt bugs, filed separately: `(. Class staticMethod)` dot-form static call, and `load-string` not applying the `#inst` reader.

Breaking: `java.time.*` is no longer available out of the box — a program using it must depend on jolt-lang/time. Core `make test` green; site updated.


yogthos:
Also adds Maven-coordinate resolution to `jolt.deps` (commit f8d9bc2). The resolver skipped `:mvn/version` coords, so tick's own transitive deps (cljc.java-time, time-literals — declared as Maven) couldn't resolve and had to be re-pinned as git deps in the time library. jolt now fetches the JAR source and reads its pom for transitive deps, staying lean by skipping non-loadable / ClojureScript-toolchain subtrees. This lets the time library depend on just tick.

PR #385, opened 2026-07-17
perf: record assoc without the triple copy; O(1) with-meta for vectors

yogthos:
Two measured hot paths from the perf audit (epic jolt-mw44):

- record assoc/dissoc built the new record via jrec-vals (fresh vector) + jrec-vec-copy (second copy) + make-jrec (apply/vector->list + rest list) — three traversals and two allocations to change one field, 137ns/op vs the JVM's 23. make-jrec-from-existing dispatches on nfields and calls the fixed-arity ctors with direct slot reads; the ^double coercion flags, extmap paths, and spill records all keep their semantics (corpus rows assert type/equality/meta/hash preservation, source immutability, and >8-field spills).

- with-meta on a vector materialized the whole trie and rebuilt it (O(n) + O(n log n)) just to give the meta side-table a fresh identity; it now shares cnt/shift/root/tail structurally — one allocation. jrec meta-copy routes through make-jrec-from-existing. Rows assert value equality on large vectors and that the original's metadata is untouched.

Implemented by a dirge session (audit worktree), reviewed, probed, and committed by the supervisor; gated on the hardened corpus (throws-asserting + crash-baselined) merged from main. A/B benches follow once the machine is quiet and will be posted here.

jolt-mw44.25, jolt-mw44.31


yogthos:
Deferred A/B/A benches (quiet machine, A=origin/main, B=this branch):

- record assoc, 1M iters: A 131 / B **92** / A 122 ns/op (~28% faster; remaining gap to the JVM's ~23ns is hasheq-cache and meta bookkeeping, in scope for jolt-mw44.32)
- with-meta on a 1000-element vector, 100k iters: A 60,690 / B **350** / A 60,680 ns/op (**~173x** — the O(n) materialize+rebuild is gone)

PR #386, opened 2026-07-17
jolt build: direct-link + whole-program inference by default

yogthos:
A plain jolt build emitted (jolt-invoke1 (var-cell-deref cell) x) for every cross-fn call and ran no cross-fn type flow — the single biggest throughput lever the perf audit identified sat behind --direct-link/--opt. Release and optimized builds now direct-link app namespaces and run wp-infer.

Escape hatches, all smoke-asserted end to end: --dev and --no-direct-link (or deps.edn :jolt/build {:direct-link false}) opt out; --direct-link remains as a redundant alias; JOLT_NO_WP_INFER=1 skips the fixpoint for very large apps. The documented tradeoff: a plain def is frozen in the built binary, ^:redef/^:dynamic defs stay var-routed so runtime redefinition and binding keep working (asserted by running the built binary).

Running the two together exposed a latent wp-cache bug that had been silently corrupting macro expanders in every --opt build: bld-wp-infer!'s walk skipped macro and analyzer-rejected forms while the emit walk consumes the per-ns IR cache positionally — from the first macro in a namespace, every form got the NEXT form's IR. Invisible before (AOT binaries never call expanders); direct-link turned it into invalid-Scheme compile failures. Skipped forms now leave #f placeholders (fresh analysis on consumption), and the contagion prepass filters them.

Smoke: plain release direct-links the app->app call; wp-infer proven at the emit level (fl-op count drops under JOLT_NO_WP_INFER=1); --no-direct-link stays var-routed; ^:redef/^:dynamic behavior verified in the built binary. Baselines confirmed against origin/main in a detached worktree. Gated on the hardened corpus merged from main.

jolt-mw44.26 (session implementation; wp-cache fix + review by supervisor)

### DIFF
--- a/host/chez/build-smoke.sh
+++ b/host/chez/build-smoke.sh
@@ -55,6 +55,47 @@ if [ "$got" != "$want" ]; then
   exit 1
 fi
 
+# --- release now defaults to direct-linking + whole-program inference ------------
+# A plain `jolt build` (release, no flags) must direct-link app->app calls (the
+# throughput lever the perf audit identified) AND run wp-infer — both were opt-in
+# (--direct-link / --opt) before. $out is still the plain release build here.
+
+# The cross-ns app.core -> app.util/shout call lowers to a direct jv$ binding in
+# the plain release build, not var-deref.
+if ! grep -q '(jv\$app.util\$shout' "$out.build/flat.ss"; then
+  echo "  FAIL: release build did not direct-link the app->app call"; exit 1
+fi
+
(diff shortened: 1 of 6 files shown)


yogthos:
Deferred A/B/A bench (quiet machine): a 10M-iteration cross-namespace call loop in a DEFAULT (flagless) build — A=origin/main 126ms / B=this branch **51ms** / A 125ms. **2.5x** on cross-fn-call-dominated code with identical output, from direct-linking + wp-infer alone; exactly the lever the perf audit measured behind --opt, now what users get by default.

PR #417, opened 2026-07-19
fix(build): read the require scan in scan mode — accept ::alias/kw keywords

### LINKED ISSUE #416: joltc build rejects alias-resolved keywords (::alias/kw) that joltc run accepts
Alias-resolved auto keywords — `::str/trim-mode` with `[clojure.string :as str]` required — work under `joltc run` and the REPL, but `joltc build` fails while scanning requires:

```
Unhandled exception: Invalid token: ::str/trim-mode
  trace:
    rdr-read-keyword
    ...
    bld-ns-requires
    dfs
    bld-require-closure
    build-binary
```

Self-namespaced `::kw` is unaffected.

**Minimal repro:** https://github.com/burinc/jolt-examples/tree/repro/build-alias-keywords/build-alias-kw — a 10-line project; the README shows the run-ok/build-fail transcript. Jolt v0.4.2, macOS AArch64.

**Root cause:** the loader reads and *evaluates* one form at a time, so the `(ns …)` form registers its aliases before later forms are read. `bld-ns-requires` (`host/chez/build.ss`) reads **all** top-level forms up front via `ei-read-all` with no ns evaluation, so `chez-resolve-alias` finds nothing and `rdr-read-keyword` throws. The emit-path reads don't hit this — they happen after loading, with aliases registered and `set-chez-ns!` in effect.

Found while porting the liquid editor (~6k LOC of idiomatic namespaced-keyword Clojure) to jolt — 411 `::alias/kw` occurrences had to be rewritten to fully-qualified keywords before the project would build. Fix proposed in the companion PR.

burinc:
Fixes the `Invalid token: ::alias/kw` failure in `joltc build`'s require scanner (companion issue has the minimal repro and analysis).

Adds a reader parameter `rdr-scan-mode` (default `#f`): in scan mode an unresolvable `::alias/name` keeps the alias text as the keyword ns instead of erroring; `bld-ns-requires` parameterizes it around its read. The placeholder value is never observed — the scanner only extracts require clauses and discards every other form — and compiled output is correct because the emit-path reads happen after loading (aliases registered, `set-chez-ns!` in effect).

Verified:
- The repro project builds, and its binary prints the correctly-resolved keyword (`{:clojure.string/trim-mode :both}`) — the placeholder never leaks into compiled output.
- A genuinely invalid `::nosuch/kw` still errors under `run`/load (scan mode is scanner-scoped).
- `make unit` green; `make corpus` green (only the 9 pre-existing allowlisted failures).

Fixes #416.

### DIFF
--- a/host/chez/build.ss
+++ b/host/chez/build.ss
@@ -612,7 +612,11 @@
                              (set! reqs (cons (car parsed) reqs)))))
                        (expand-spec unquoted))))
                   (cdr items)))))))
-      (map rdr-form->data (ei-read-all src)))
+      ;; scan mode: this read happens BEFORE any namespace is loaded, so
+      ;; alias-resolved auto keywords (::alias/kw) can't resolve yet — read
+      ;; them leniently; only require clauses are extracted from these forms.
+      (parameterize ((rdr-scan-mode #t))
+        (map rdr-form->data (ei-read-all src))))
     (reverse reqs)))
 
 ;; Post-order DFS from a list of root namespace names: for each name, find its
--- a/host/chez/reader.ss
+++ b/host/chez/reader.ss
@@ -65,6 +65,12 @@
 ;; its tagged elements through :readers/:default like the JVM.
 (define rdr-edn-mode (make-parameter #f))
 (define rdr-discard-cb (make-parameter #f))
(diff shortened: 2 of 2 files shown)


yogthos:
and new release is out with the fix

PR #424, opened 2026-07-19
Run flatland/ordered on jolt: deftype collection interop + print-method

yogthos:
flatland.ordered builds `OrderedMap`/`OrderedSet` as deftypes over the `clojure.lang.*` interfaces, which shook out a cluster of general gaps. Its suite now passes **129/129**. All runtime .ss (no re-mint); full `make test` green, +1 JVM-certified corpus row.

## General fixes
- **Native-collection Java interop** — `.cons`/`.assoc`/`.without`/`.assocN`/`.disjoin`/`.pop`/`.asTransient`/`.hashCode` and the transient `.conj`/`.assoc`/`.without`/`.persistent`/… dispatch on a vector/map/set (and a jolt-transient), so a deftype calling these on its backing collection works.
- **conj/contains?/disj/get/transient/persistent!/hash** route to a deftype's own `cons`/`containsKey`/`disjoin`/`get`/`asTransient`/`persistent`/`hasheq` method. The generic jrec `conj`/`contains?` arms are guarded so a deftype's own method wins (they were registered later and shadowed it).
- **print-method** — `pr`/`pr-str` render a record through a user `(defmethod print-method SomeType …)`. An unqualified `defmethod` in a required library now extends `clojure.core/print-method` (implicit `refer-clojure`) instead of a dead per-ns shadow.
- **instance?** on a deftype walks its declared interfaces' ancestry (an `IPersistentMap` is a `Counted`/`Associative`/`IPersistentCollection`); `register-inline-protocol!` grafts an imported interface under its canonical FQN.
- **Collection .hashCode** is the `java.util.Map`/`Set`/`List` hashCode (was Murmur3 hasheq); `APersistentMap/mapHash` shimmed; `hash-unordered` wraps at 32 bits so a `hash-unordered-coll` type hashes equal to a plain map/set.
- A bare imported deftype class name resolves to its class value, `= (type inst)`; a flat `(:import a.b.Type)` binds it.


yogthos:
Follow-up from a dirge review: fixed a real bug — the guarded plain `contains?` arm excluded only `containsKey`, not `contains`, so `(contains? ordered-set member)` returned false (the field-presence arm shadowed the set's `contains` method). Guard now skips a type declaring either; regression row added. Full `make test` green.

PR #427, opened 2026-07-20
Download Maven deps over jolt-native HTTPS; drop curl

yogthos:
Replaces the `curl` shell-out in `jolt.deps`'s Maven download with a jolt-native, cert-verifying HTTPS GET. `joltc` now downloads dependencies itself — no external `curl` — and does it safely (TLS with peer verification), instead of over an unauthenticated transport.

## What
- **`stdlib/jolt/mvn_http.clj`** — a minimal HTTPS GET-to-file: raw BSD socket via `jolt.ffi`, TLS over the system **OpenSSL** through in-memory BIOs (`SSL_CTX_set_default_verify_paths` + `VERIFY_PEER` + SNI + `SSL_set1_host` hostname check), HTTP/1.1 with redirect following, binary-faithful body written to disk. Adapted from jolt's own proven `net.clj`/`tls.clj` (jolt-lang/http-client) — **not** a dependency on that library (chicken-and-egg), a minimal copy in core using `jolt.ffi`.
- **`jolt.deps`** — `ensure-maven` calls `jolt.mvn-http/fetch`; **all curl removed**.
- OpenSSL is **lazy-loaded** only on the first fetch; a box without it (or a repo that can't be reached) fails the fetch gracefully — non-fatal, resolution continues.

## Why not plain HTTP / a vendored TLS lib
Plain HTTP is unsafe for executable jars. The one vendorable pure-Chez TLS binding (`ecraven/chez-scheme-libraries`) is GPL-3, and pure-Scheme TLS (Industria) is too old for modern repos. OpenSSL via FFI is a shared *library* (like libsqlite3/libz jolt already links), not an external *program*.

## Platforms
- **macOS / Linux**: implemented and **validated** — live fetches of `.pom` and a binary `.jar` (valid `PK` header) from Central and Clojars, graceful 404, and a fresh-repo end-to-end resolve of `org.clojure/data.json`. On macOS it loads the **Homebrew** OpenSSL; the bare `libcrypto.dylib` name is deliberately excluded because the protected system copy aborts the process (uncatchable).
- **Windows**: Winsock path (`ws2_32` + `WSAStartup` + `closesocket`, Win64 offsets) is implemented but **NOT yet validated on a Windows host** — no Windows machine was available. Needs a real-Windows smoke before it's trusted.

## Not in scope
`git` (git deps) and `unzip` (jar extraction) still shell out — separate, larger efforts (git smart-HTTP protocol; a jolt-native ZIP reader).

## Test
Opt-in `make httpsfetch` smoke (network; Central + Clojars). Full `make test` green (selfhost fixpoint, corpus 0-new, unit 1002/1002, cts baseline, dce-refs, shake, manifest).

## Known limitations
No socket read-timeout (a stalled server hangs a fetch); whole response buffered in memory (fine for jars).

Implemented by dirge from a written spec, then reviewed — review caught the macOS abort risk and the missing Windows candidates, and added the Winsock path. Docs (README + site) updated with the per-platform OpenSSL/git/unzip requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Maven dependencies download over HTTPS through jolt itself instead of shelling
+  out to `curl`. A new `jolt.mvn-http` does a minimal cert-verifying HTTPS GET
+  (a raw socket via `jolt.ffi`, TLS over the system OpenSSL with peer verification
+  + SNI + hostname check), so a dependency jar is never fetched over an
+  unauthenticated transport. It loads the real OpenSSL lazily — on macOS the
+  Homebrew `openssl@3` (the protected `/usr/lib` copy can't be loaded); on Linux
+  the distro `libssl`/`libcrypto`; on Windows the `libssl-3-x64.dll` DLLs via
+  Winsock (the Windows path is implemented but not yet validated on a Windows
+  host). A repo that can't be reached is skipped, non-fatal. `git` (git deps) and
+  `unzip` (jar extraction) are still shelled out; `curl` is gone.
+
 ## [0.4.7] - 2026-07-19
 
 Library conformance (flatland/ordered; babashka.http-client via jolt-lang/http-client),
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 CHEZ ?= $(shell command -v chez 2>/dev/null || command -v chezscheme 2>/dev/null || command -v scheme 2>/dev/null)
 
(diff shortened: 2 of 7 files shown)


yogthos:
Security-focused dirge review done. Verdict: **TLS verification itself is sound** — verify-peer + default paths + SNI + `SSL_set1_host`, fail-closed on a bad cert (no MITM path). Fixed the robustness follow-ups it found:

- **read/send timeout** (SO_RCVTIMEO 30s) — a stalled/malicious repo could otherwise wedge resolution forever in a blocking `recv`
- **OpenSSL object leak** when `connect` throws (DNS-fail/refused — the common path): open the socket first, hand BIOs to the SSL before any fallible step, free via `tls-close` on error
- **truncation guard**: verify body length vs Content-Length + temp-file + atomic rename, so a mid-transfer drop can't leave a corrupt jar deps.clj would trust
- **CRLF/NUL request-smuggling guard**; explicit **http:// downgrade refusal** in `resolve-location`; follow **303**; pin **TLS ≥ 1.2**; assert verify-setup returns

Added **pure-function tests** (`test/mvn_http_test.clj`, wired into the gate as `make mvnhttp` — no network/OpenSSL): parse-url, resolve-location incl. the downgrade refusal, header-end/ci, dechunk (binary-exact), parse-response framing. Full `make test` green (incl. mvn-http-test, selfhost fixpoint, certify 0-new/0-stale). Re-validated the live path on macOS.

PR #431, opened 2026-07-20
Move java.time.* out of core into jolt-lang/time (RFC 0008)

yogthos:
Implements [RFC 0008](https://github.com/jolt-lang/jolt-lang.github.io/blob/main/resources/md/rfc/0008-time-core-library-split.md).

## The problem

Core registered a partial `java.time` surface next to the `#inst` literal, so some `java.time` calls worked with no dependency and some didn't, with no rule a user could predict:

```clojure
(java.time.Instant/now)        ; worked  — core registered it
(java.time.LocalDateTime/now)  ; worked  — core registered it
(java.time.LocalDate/now)      ; failed  — library only
(java.time.ZonedDateTime/now)  ; failed  — library only
```

This is the confusion reported in the field. The two that worked were the anomaly.

## The change

Core now keeps only what the language itself needs: the `#inst` reader literal and the `java.util` / `java.text` date layer it round-trips and formats (`Date`, `sql.Date`/`Timestamp`, `Calendar`, `TimeZone`, `SimpleDateFormat`, `Locale`). The whole `java.time.*` namespace is the jolt-lang/time library.

The library already registers the full `java.time` surface through the same host seams and overrides core, so there is nothing new to write on the library side. Removing core's copies also ends the duplicate-registration drift between the two (the warnings that showed up under `JOLT_DEBUG`).

Core produces no `java.time` value of its own now. The `Date`/`FileTime` `.toInstant` bridge routes through the library's `Instant` constructor via the existing `set-instant-ctor!` hook; with no library loaded it throws a message naming the dependency. An unresolved `java.time.*` class now reports that it lives in jolt-lang/time rather than a bare "Unknown class", so the fix travels with the error.

```
$ joltc -e '(java.time.Instant/now)'
java.time.Instant requires the jolt-lang/time library — java.time is not in core (add io.github.jolt-lang/time to deps.edn)
```

The non-standard `Date.toLocalDate` / `toLocalDateTime` that core had invented (a `java.util.Date` has no such methods on the JVM) are dropped.

## Verification

All against the modified core:

- `make unit` 994/994 (insttime 26/26, including a new row asserting the bridge requires the library)
- `make selfhost` fixpoint holds
- `make smoke` 75/75, `make ffi` 10/10 (the HTTP date path uses `SimpleDateFormat`, which stays)
- `make corpus` (only the pre-existing allowlisted failures)
- the jolt-lang/time gate: 97 tests, 1020 assertions, 0 failures

`host/chez/*.ss` load at runtime rather than baking into the seed, so there is no re-mint and the checked-in seed is unchanged.

## Compatibility

A program that called `java.time.Instant/now` (or formatted with `DateTimeFormatter`) with no dependency now needs jolt-lang/time. That is the point: the boundary is consistent. The library provides a strict superset, and a program formatting a `Date` without `java.time` can use `java.text.SimpleDateFormat`, which stays in core.

### DIFF
--- a/host/chez/java/host-static.ss
+++ b/host/chez/java/host-static.ss
@@ -192,6 +192,23 @@
 (vector-set! (mutable-static-cell "clojure.lang.RT" "checkSpecAsserts" #t) 0 #f)
 
 ;; ---- emit entry points ------------------------------------------------------
+;; An unresolved java.time.* class (fully qualified, or a distinctive short name)
+;; almost always means the jolt-lang/time library isn't loaded: java.time lives
+;; there, not in core (RFC 0008). Name the dependency so the fix is obvious,
+;; rather than leaving a bare "Unknown class".
+(define jt-short-names
+  '("Instant" "LocalDate" "LocalTime" "LocalDateTime" "ZonedDateTime"
+    "OffsetDateTime" "OffsetTime" "ZoneId" "ZoneOffset" "Duration" "Period"
+    "YearMonth" "MonthDay" "DateTimeFormatter" "FormatStyle"))
+(define (java-time-class? class)
+  (or (and (>= (string-length class) 10) (string=? (substring class 0 10) "java.time."))
+      (and (member class jt-short-names) #t)))
+(define (unknown-class-message class)
+  (if (java-time-class? class)
+      (string-append class " requires the jolt-lang/time library — java.time is not in "
+                     "core (add io.github.jolt-lang/time to deps.edn)")
+      (string-append "Unknown class " class)))
+
 (define (host-static-ref class member)
   (let ((cell (mutable-static-cell class member #f)))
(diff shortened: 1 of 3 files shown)


yogthos:
Superseding this. After surveying the conformance libraries, the boundary is changing: instead of removing java.time from core entirely (which made basic things like Instant/now require the dependency), core will carry a base java.time API (Instant, Local*, Duration, Period, Year/YearMonth, ZoneOffset, ZoneId construction, DateTimeFormatter patterns) and the library keeps only named-zone rules, ZonedDateTime/OffsetDateTime, localized formatting, and tick. The dedupe is done by moving the library's existing Clojure base into core rather than reintroducing a parallel Scheme copy. See the revised RFC 0008. A fresh PR will implement the new boundary.

PR #482, opened 2026-07-27
Drive the build gates through the prebuilt binary

yogthos:
`make -j8` over the whole gate goes from 338s to 124s on this machine, about 2.7x.

## What was slow

I timed every target in `make ci` individually. Five of them accounted for 73% of a serial run, and all five for the same reason: they shell out to `bin/jolt` in source mode. A `jolt build` costs about 12.5s that way and about 2.5s through `target/release/jolt`, and `buildsmoke` alone drives 26 builds. `smoke` and `cts` already took `testbin` and defaulted `JOLT_BIN` to the prebuilt binary; the build-oriented gates never got the same treatment.

Measured with a warm `testbin`:

| gate | before | after |
|---|---|---|
| `buildsmoke` | 343.6s | 92.1s |
| `shakelocal` | 204.4s | 42.0s |
| `depssmoke` | 120.7s | 7.7s |
| `staticnativesmoke` | 29.9s | 6.6s |
| `buildlibsmoke` | 14.9s | 12.6s |
| total | 713.5s | 161.0s |

This matters more than the totals suggest because CI runs `make -j$(nproc)`, where wall time is bounded below by the longest single gate. `buildsmoke` at 5.7 min was the critical path; at 1.5 min it no longer is.

`JOLT_BIN=bin/jolt` puts any of them back in script mode.

## Coverage

Switching means `buildsmoke` stops exercising the source-mode build driver, which is what a developer actually runs. `devbootsmoke` case (e) already builds a project through the devcache launcher and runs the result, so it was not going to be uncovered, but relying on that alone felt thin for a gate whose whole subject is `jolt build`. So `buildsmoke` keeps one explicit `bin/jolt` build at the end over the small `jolt-ext-app` fixture, mirroring how `smoke.sh` keeps a script-mode case. It costs 12.5s and is skipped when `JOLT_BIN` already is `bin/jolt`.

## testbin

Adding the prerequisite exposed a second problem. `testbin` rebuilt unconditionally, which is free under `make -j ci` where it is one shared node in the graph, but it charged every single-gate run 18s. That was enough to make `make buildlibsmoke` slower with the new prerequisite than without it, 14.9s before against 30s after. It now rebuilds only when a file under `host/chez`, `jolt-core`, `stdlib` or the vendored source roots is newer than the binary, which is the same set `build-jolt.ss` bakes in. `JOLT_FORCE_TESTBIN=1` rebuilds anyway. This also speeds up single runs of `smoke`, `cts` and `aotcachesmoke`, which were already paying it.

## Testing

Ran the full gate in parallel before and after, cold with no prebuilt binary, everything passing both times: 338s to 124s. Also ran each of the five individually and confirmed the assertion counts are unchanged (`deps-alias smoke: 47 passed`, `cli smoke: 87 passed`, `cts: 5717 pass`, `build smoke: passed`, `shake smoke: passed`).

`sci` is excluded from those runs. It fails in my checkout at 209 against a 211 floor from vendored submodule drift, unrelated to this change and unaffected by it.

## Left on the table

Twenty pass gates (`infer`, `wp`, `pic`, `narrow`, and so on) take about 1.5s each, essentially all of it `run-gate-harness.ss` loading the runtime from source. That is 30s aggregate and near zero under `-j`, but it makes iterating on a single pass gate sluggish. A precompiled preamble would take each to about 0.2s. It needs its own artifact rather than reusing `target/dev/flat.so`, because the gates deliberately stop at `compile-eval.ss` and do not load `loader.ss`. Not done here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **The gate runs about 2.7x faster.** `make -j8 ci` drops from 338s to 124s.
+  The build-driving gates (`buildsmoke`, `shakelocal`, `depssmoke`,
+  `staticnativesmoke`, `buildlibsmoke`) shelled out to source-mode `bin/jolt`,
+  where a `jolt build` costs ~12.5s against ~2.5s through a prebuilt binary, and
+  `buildsmoke` alone drives 26 of them. They now take `testbin` and default
+  `JOLT_BIN` to `target/release/jolt`, the way `smoke` and `cts` already did:
+  those five together go from 713s to 161s. `buildsmoke` keeps one explicit
+  `bin/jolt` build at the end so the source-mode driver stays gated, and
+  `JOLT_BIN=bin/jolt` puts any of them back in script mode. `testbin` itself is
+  now rebuilt only when something it bakes in is newer than the binary —
+  unconditional rebuilds were free under `make -j ci` but charged every
+  single-gate run 18s, enough to make `make buildlibsmoke` slower with the new
+  prerequisite than without it.
+
 ## [0.5.7] - 2026-07-27
 
 A `.jolt` source extension, `-e` and `:main-opts` fixes across the CLI entry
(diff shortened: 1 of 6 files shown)


yogthos:
Rebased onto main after the 0.5.7 cut. The conflict was only the changelog entry, which the release moved out from under; it sits under `[Unreleased]` again. The Makefile auto-merged cleanly against the new `oparity` gate.

Re-ran the full gate on the rebased branch, cold with no prebuilt binary: 131s, everything green, including the retained `source-mode-driver` case in build smoke.

PR #488, opened 2026-07-28
Fix review findings: inline-pass throw elision, bigdec arm registration

yogthos:
Fixes from a focused correctness/architecture review of the compiler and host runtime.

**Correctness**

- The inline pass admitted `:throw` via `safe-op?`, so `pure?`/`total?` treated it as discardable and `elim-let-structs` dropped const-keyed map bindings with a throwing unread value. An optimized release build folded `(let [m {:a 1 :b (throw "boom")}] (:a m))` to `1` and swallowed the exception. `:throw` is now relocatable (still `safe-op?`) but never pure/total. Regression case in `test/chez/inline-throw-app` plus a build-smoke check; seed image re-minted.
- Deletes the dead `all-vals-pure?` and fixes the stale `natives-num.ss` header (numbers aren't all doubles; bit ops return exact wrapped integers).

**Architecture**

- `post-prelude.ss`'s `seqable?` patch no longer hardcodes the java shim tag list; it consults `jhost-seqable-shim?` in `java/host-static-classes.ss`, next to the class table that owns that knowledge.
- The global `error` shadow and `cond-expand` shim moved from `regex.ss` to the top of `rt.ss` (every load path enters there), so nothing before `regex.ss` sees un-normalized `error`.
- `bigdec.ss` no longer `set!`-mutates seventeen core vars at load time. `seq.ss` owns a new `register-num-arm!` extension point (same convention as `register-eq-arm!`/`register-compare-arm!`); bigdec registers its arithmetic, predicate, cast, compare, and `==` tower arms through it. The core's knowledge of the jbigdec representation in `predicates.ss`/`converters.ss`/`natives-seq.ss` is gone.

**Verification**

- `make inline`, `make unit`, `make smoke`, `buildsmoke` (including the new non-zero-exit inline-throw check), and `manifestcheck` all pass.
- Corpus gate: 3839/3858 pass, 0 NEW divergences, 173 throws verified. The 10 informational crashes are identical to the base commit (label-for-label, verified via `JOLT_DUMP_CRASH_LABELS=1` on `bc20ecea`) — none introduced here.
- BigDec behavior spot-checked by hand: arithmetic contagion, `(== 3M 3)` vs `(= 3M 3)`, `compare`/`sort`/`min`/`max` across bigdec/long/double, and error paths unchanged.


yogthos:
Reviewed and verified. One regression found and fixed on the branch.

The `num-equiv-slow` extraction broke multi-arg `==` with a bigdec operand. The pairwise loop in `jolt-num-equiv` runs as soon as *any* operand is a shim number, so an individual pair inside it can still be two plain numbers. bigdec's arm declines those to the base, and the new base threw unconditionally, so `(== 1 1 1M)` and `(== 1 2 3M)` raised ClassCastException where main answers true and false. The base now handles plain-number pairs with the same inexact-contaminates rule the all-plain branch uses, and throws only for a non-number. Also made bigdec's `decimal?` arm decline to `prev` instead of dropping it, since `register-num-arm!` documents that as the contract and the next numeric shim would have been clobbered by load order.

The rest checks out. `jhost-seqable-shim?` is faithful to the tag list it replaces and is bound unconditionally from rt.ss before post-prelude runs. The rt.ss preamble move is safe: nothing in the host calls `error` with a string first argument, so no call site changes shape, and `%chez-error` is the only external reference. All nineteen op names in `register-num-arm!` resolve to real top-level vars. Dropping bigdec's `def-var!` of `compare` is correct because `jolt-compare` walks the arm list at call time. The natives-num.ss header now matches what the bit ops actually do.

I confirmed the inline-pass bug is real: built the new test app with main's compiler under `--opt --direct-link` and it printed `1` instead of `THROW2 OK`.

Verification after merging main in: full `make ci` passes twice, corpus 0 NEW divergences with 173 throws verified, cts matches baseline, SCI 211/218, buildsmoke covers the new THROW2 case. `make remint` converges in one pass with no seed diff. Also diffed 60 bigdec, seqable, bit-op and `==` cases against main output and they are identical.

PR #500, opened 2026-07-29
Use Makes for Jolt build workflow

ingydotnet:


yogthos:
Held this out of 0.5.13 — it breaks the Linux release build. Two findings, one blocking.

## Blocking: Makes provisions its own Chez on Linux, without `--threads`

`ci/glibc-floor-build.sh` builds Chez 10.4.1 with `--threads` into a cached
`/opt/chez`, puts a `chez` wrapper on PATH, then runs `make jolt-release` inside
`quay.io/pypa/manylinux2014_x86_64`. With this branch, Makes ignores that Chez —
`CHEZSCHEME-LOCAL` is set unconditionally on Linux — and provisions its own:

```
$ podman run --rm --platform linux/amd64 quay.io/pypa/manylinux2014_x86_64 …
which scheme: /opt/chez/bin/scheme; version: Chez Scheme Version 10.4.1
CHEZSCHEME-LOCAL := /work/.cache/local/chezscheme-10.4.1
CHEZSCHEME       := /work/.cache/local/chezscheme-10.4.1/bin/scheme
```

with a real `scheme`, `chez` and `petite` 10.4.1 on PATH. The configure line it
uses:

```
./configure \
  --installprefix=/work/.cache/local/chezscheme-10.4.1 \
  --disable-x11 \
  --disable-curses \
  CC=/work/.cache/local/gcc-14.2.0-1/bin/gcc
```

No `--threads`. `host/chez/java/concurrency.ss` says so directly ("Requires a
threaded Chez build (fork-thread / make-mutex / make-condition)"), and it uses
`fork-thread` unconditionally for agents, futures, `pmap`, and the `:blocking` FFI
collect-safety. `tests.yml`, `release.yml` and `ci/glibc-floor-build.sh` all pass
`--threads` today.

Secondary on the same path: it downloads xPack GCC 14.2.0 and builds with that
instead of the container's toolchain. The container is manylinux2014 specifically to
hold the glibc floor at 2.17 (#455/#452); a 14.2.0 toolchain from outside it likely
raises the floor back. The `STATIC_DEPS=1` static ncurses/tinfo/zlib linkage in
`glibc-floor-build.sh` is also built around the Chez that script installs, and
`--disable-curses` changes it.

macOS is fine — there Makes accepts the Homebrew Chez (`CHEZSCHEME := /opt/homebrew/bin/chez`),
which is why `tests.yml` is green. The release workflow is never exercised by a PR,
so CI passing here doesn't cover it.

Fix shape, whichever you prefer: have the Linux path honor an existing
`CHEZ`/`CHEZSCHEME` (or `scheme` on PATH) the way macOS does, or add `--threads` to
Makes' configure and keep the container's CC. Worth checking the produced binary's
glibc floor with the `glibc-floor.yml` probe either way.

## Minor: `.cache/` isn't gitignored

The parse-time `git clone` into `.cache/makes` leaves an untracked nested repo after
any `make`:

```
$ git status --short
?? .cache/
```

`git add -A` after a build then tries to add a nested repo. One line in
`.gitignore`.

## Non-issues I checked

- GNU Make 3.82 in manylinux2014 parses the new Makefile fine.
- `JOLT-TARGETS-NEEDING-DEPS` covers every target that touches Chez or `bin/jolt`;
  the ones left out (`certify`, `manifestcheck`, `build`, `install`, `default`) are
  either pure text processing, JVM-only, or reach `deps` through `testbin`.
- `.PHONY: build install test ci` covers the four names that collide with a
  file/directory in the tree (`build/`, `target/`, `test/`, `ci/`, and the tracked
  `install` script).
- `make clean`'s `MAKES-CLEAN := build/ target/` only removes gitignored products.

PR #528, opened 2026-08-03
Fix the host gaps found running libraries' own test harnesses

yogthos:
Replaying third-party suites through each library's real runner instead of the bespoke `conformance-run` harness surfaced a batch of host bugs the old recipes structurally could not reach — they hand-listed namespaces and supplied their own dep sets, so whole code paths never ran.

The largest single find is the protocol identity bug; everything else is a gap one of the suites walked into.

### Protocol identity

Protocol dispatch keyed a method table by the protocol's SIMPLE name, so two protocols named alike in different namespaces shared one table and the later `extend` silently replaced the earlier one — a wrong answer, not a crash. Dispatch now keys by `"<defining-ns>/<Name>"`, resolved through `:refer` and `:as` by a new `jolt.host/protocol-key-of`. A symbol naming no protocol (`Object`, `java.util.Map`, an imported `clojure.lang.ILookup`) keeps its bare name, because that is how `value-host-tags` spells the tags a value reports.

This is what was holding malli down: 1,874 → 16,280 assertions in its own suite, ClassCastException 594 → 1. No dispatch cost — `dispatch` 69.9/66.9/69.9 ms A/B/A.

### Host classes and core

- `java.io.File` compares by pathname under `=`. `.equals` and `hash` already agreed, so two Files built from one path were unequal only through `=`, which is how ring-core's resource tests read them. The two-arg constructor also joins with exactly one separator; a parent ending in `/` produced `a//c`.
- `(char n)` spans the Unicode scalar values. jolt's strings are code-point indexed, so it could not rebuild a char `(first s)` had just handed out, and data.json errored writing an astral character. This is a deliberate superset of the JVM's 16-bit char — the clojure-test-suite already treats that assertion as host-dependent (jank and basilisp behave as jolt now does), so it lands as a cts baseline line plus a `:string-model` divergence entry, not as parity.
- `(. Class MEMBER)` with no arguments reads a static FIELD when one is registered. It always emitted a static call, so a field's value was applied as a zero-arg procedure and came back **nil** — a silent wrong answer. `(. Class -MEMBER)` was rejected outright. Four malli namespaces load-failed on `LocalDate/-MIN`. (bead jolt-c0ri)
- Added `System/arraycopy`, with the JVM's overlapping-copy semantics.
- A `reify` declaring `IFn` is `ifn?` and an instance of it.
- Protocol dispatch picks the most specific class; added `clojure.datafy`, `Murmur3/hashCombine`, `clojure.java.javadoc`; fixed cross-namespace FQ-name `extend`.
- iconv now issues the POSIX reset call, so a stateful charset emits its closing escape. `(.getBytes "い" "ISO-2022-JP")` was 5 bytes where the JVM gives 8 — RFC 1468 requires the text to end in ASCII.
- `LineNumberingPushbackReader` gets its own jhost tag so it reports its own class. tools.reader does `(extend LineNumberingPushbackReader IndexingReader …)`, which never dispatched.
- `io/resource` answers an absolute `file:` URL. Source roots are relative, and resolving a name against `file:./test/x` throws `MalformedURLException: no protocol`.
- Added `Character/codePointAt` and Runtime's memory API over Chez's heap accounting.
- `(. Foo bar x)` on an imported simple name is a static call, so a method the class token does not answer routes through `host-static-ref`, where the on-demand autoload lives. (bead jolt-5hnv)

### clojure.test reports like reference Clojure

A failure printed one line. The reference prints a blank line, `FAIL in (test-name) (file:line)`, the testing context and message, then `expected:` / `  actual:` — the shape every editor integration, CI parser and third-party reporter keys off, and which test.check's own suite asserts on directly. Verified byte-identical to 1.12.5 for a failure; an `:error` differs only in the frames underneath, which tail calls leave nothing to report.

Position comes from the reader — `(meta &form)` plus `*file*` — not a stack walk. `*report-counters*` is incremented too, including `:test` per test var.

Also fixed here: the registry runner never bound `*testing-vars*`, so a failure in a registry-run test had an empty name in the header.

### deps resolution

- Accept the sha an annotated tag carries as well as the commit it peels to. `git ls-remote` prints the tag object for `refs/tags/X`, so that is what a coordinate written from its output pins (cognitect-labs/test-runner v0.5.0 is one); tools.deps takes either. Verified against the real CLI. The tag cache is now two tokens, and a one-token file from an older jolt is re-resolved rather than trusted.
- Accept the legacy `:sha` / `:tag` spellings alongside the namespaced keys. malli pins spec-alpha2 that way, which blocked its whole test alias.
- Carry Clojure's spec deps transitively.

### stdlib

- Added `clojure.stacktrace`, which Clojure ships and jolt did not. Frame lists are empty here (tail calls, already a recorded divergence); the throwable line, ex-data and cause chain match exactly.

### Conformance shims

Three suites test against Java libraries jolt has no equivalent for. Each shim lives on a jolt-owned path under `test/conformance/libs/shims/` and goes on the recipe — never written into an upstream checkout, where the edit would be invisible once its `.git` is gone and the recorded tally would silently measure our own source: commons-codec Base64 for data.codec (RFC 4648, as the suite's oracle), commons-lang `StringEscapeUtils` for markdown-clj, commons-io `IOUtils`/`FileUtils` for ring-core.

Also fixed markdown-clj's own finding: an unparseable numeric entity now leaves its token alone.

### Effect on library tallies

Passing assertions, and what is left failing. Every figure is a `make
libconformance` measurement, not an estimate.

| lib | before | after |
| --- | --- | --- |
| malli | 121 pass, 20 load-fail | **11,934** pass, 8 load-fail |
| honeysql | 840 pass, 1 load-fail | **2,842** pass, 0 |
| rewrite-clj | 3,380 pass, 0 fail | **3,381** pass, 0 fail |
| tick | 620 pass, 1 load-fail | **723** pass, 0 |
| Selmer | 525 pass, 1 error | **526** pass, **0** |
| ring-core | 405 pass, 15 fail | **419** pass, **7** fail |
| test.check | 230 pass, 15 fail | **236** pass, **10** fail |
| tools.logging | 219 pass, 4 error | **226** pass, **0** |
| data.json | 320 pass, 2 error | **322** pass, **0** |
| markdown-clj | 180 pass, 2 error | **182** pass, **0** |
| test.chuck | 98 pass, 22 fail | **110** pass, **17** fail |
| hiccup | 385 pass, 1 fail | 386 pass, **0** fail |
| data.codec | 1 pass, 11 error | **12** pass, **0** |

malli's `fail`/`error` also rise, because seven namespaces that used to
load-fail now run at all. rewrite-clj was already complete before this branch —
163 tests / 3,381 assertions is the whole suite, the same counts reference
Clojure runs for it, and the one assertion is a real fix rather than newly
reached coverage.

### proxy extends a concrete class by delegation

`proxy` desugared to `reify`, which has no base, so a proxy over a concrete class answered only the methods its body declared — every inherited method threw — and `proxy-super` threw unconditionally. Since jolt generates no classes, a proxy now constructs a real base instance from the first super when that names a constructible class, answers what the body declares, and forwards the rest through the full dispatcher. `proxy-super` calls the base's own implementation. A proxy is an instance of its base and reports the base's class and host tags, so `instance?`, `class` and `extend-protocol` see through it. A super naming an interface has nothing to construct and stays the reify it was.

Delegation is not subclassing in one respect, recorded in `known-divergences`: the base holds no reference back, so a base method calling an overridden method runs the base's version where the JVM re-enters the override. Overriding a leaf method — the common case — is identical.

Three things had to follow, and two were bugs that predate this work. `PrintStream` did not exist (the comment beside `System/out` said `setOut` was being withheld until proxy-over-a-host-class existed, so `setOut`/`setErr` land here too). `OutputStreamWriter` transcoded the wrapped stream's port, and R6RS `transcoded-port` closes it, so `(.toString baos)` after wrapping one failed on a closed port. And `println`/`prn` never flushed although `*flush-on-newline*` existed and defaulted true with nothing reading it, while `clojure.core/flush` ignored `*out*` entirely — both now match Clojure, at 1.02–1.07x on a 200k-line println microbenchmark (1350/1413ms → 1444/1442ms).

### A URLStreamHandler decides what its URL reads

`java.net.URL` took only `(URL. spec)`, so `(URL. nil spec handler)` passed nil as the spec and raised `no protocol: `. The constructors are now told apart by argument type the way the JVM's overloads are, and a URL built with a handler reads through it whatever its protocol — `openConnection` is the handler's, `openStream` is that connection's `getInputStream`, and `slurp`/`io/reader` route the same way.

Two pre-existing bugs alongside: `getPath`/`getFile` stripped only a `file:` prefix, so every other URL reported its whole spec as its path (`(.getFile (URL. "http://h/p"))` was `"http://h/p"`, the JVM says `"/p"`); and `io/reader` on a URL read the spec as a local file path, so a non-file URL failed as a missing file rather than saying the protocol has no input.

### An incomplete gate run can no longer read as a pass

`make` stops at the first failing target, so everything after it never runs, and re-running a hand-picked subset prints per-target success that looks exactly like the whole gate's. That is how a missing host-manifest line reached CI in this branch.

The gate now runs as a sub-make behind a wrapper: a verdict line is printed either way, so the log can never end on some passing target's output — which also covers `make test | tail`, where the pipe takes tail's exit status. On a complete pass it writes a receipt naming the gate and a hash of every tracked and untracked-but-not-ignored file, and `make gate-status` answers whether *this* working tree is covered. `-i` and `-k` are refused for gate targets, at parse time, because `-i` ignores a recipe line's failure including a guard's own.

### Documented rather than fixed

ring-core's 12 remaining errors need an RFC 7578 multipart parser — library-sized, and shimming it would have ring's own multipart tests exercising our parser. data.json's 2 remaining failures are the code-point vs UTF-16 string model. tools.namespace's 1 error is `java.lang.reflect`. lasertag's 12 are the class model. Each is stated in the manifest comment for the library it affects; the model-level ones are in `known-divergences.edn`.

### Coverage

rewrite-clj keeps a `test-isolated/` tree that upstream runs as its own kaocha profile, and the isolation is the assertion: it checks a coercion works when `rewrite-clj.zip` is the ONLY namespace required, so requiring any other test namespace pulls in `rewrite-clj.node` and it passes for the wrong reason. The harness runs one process per manifest entry, so `rewrite-clj-isolated` is exactly that isolation. 1/1; rewrite-clj's own 163/3381 is unchanged, and matches what reference Clojure runs for the same suite.

### Gates

Corpus rows for every fix that is a single expression, certified against real JVM Clojure 1.12.5. A `clojure.stacktrace` acceptance test and a `clojure.test` output-shape regression, both wired into smoke.sh. 8 deps-alias checks.

`make test` green: selfhost fixpoint holds, corpus 0 new divergences, unit 1226/1226, cts at baseline, smoke 121/0, certify 0 NEW / 0 stale.


yogthos:
Pushed two more fixes from continuing the sweep (a7bc78f).

**spec.alpha is transitive from `org.clojure/clojure`.** The coordinate is intrinsic — jolt *is* Clojure, and putting the artifact's source on the roots would shadow core — but that does not make its *dependencies* intrinsic. spec.alpha and core.specs.alpha are part of core on neither host, so on the JVM a project declaring only Clojure still gets `clojure.spec.alpha`, and libraries lean on that without ever naming it (kaocha, spec.alpha's own suite). Dropping the coordinate whole took its children with it.

Versions come from the declared Clojure's own POM, so they match what tools.deps resolves:

| declared | jolt | JVM |
| --- | --- | --- |
| clojure 1.12.4 | spec.alpha 0.5.238, core.specs.alpha 0.4.74 | same |
| clojure 1.9.0 | spec.alpha 0.1.143, core.specs.alpha 0.1.24 | same |

A pinned pair covers an unreadable POM, and that lookup is quiet — it is not degrading to a jar's `pom.xml` the way the existing warning says, since jolt never loads Clojure's jar.

**A library class autoloads by imported simple name.** `host-new` already attempted the autoload, but `jt-library-names` is hand-maintained and had drifted from what jolt-lang/time registers. `DateTimeFormatterBuilder` was absent, so the fully-qualified form autoloaded (it reaches the autoload via `java-time-prefixed?` regardless) while `(DateTimeFormatterBuilder.)` after an import did not — which is how libraries actually write it. It kept `malli.transform` from loading.

The grenadine test that asserted "Clojure is pruned" now asserts what it was actually protecting: Clojure itself absent, its spec children present.

`make test` green, certify clean: 41 of 41 divergences known, 0 NEW, 0 stale. The spec.alpha entry is removed from `known-divergences.edn` — it is no longer a divergence.

Paired with jolt-lang/time#5, which fills in the builder itself, malli goes from not loading at all to 199 tests / 1865 assertions.

PR #542, opened 2026-08-06
feat: POSIX FFI bindings for java.net.Socket, ServerSocket, InetSocketAddress

allen-munsch:
This PR adds runtime POSIX FFI bindings for three Java socket classes. They register at runtime via `__register-class-ctor!` and `__register-class-methods!` using `jolt.ffi` — no Jolt rebuild needed.

## What it enables

I have been using this in **rt-voice-wasm** (a real-time speech-to-text project) in two places:

### 1. Native HTTP server — `web/serve.jolt`

Straightforward `java.net.ServerSocket` → `accept` → read HTTP request → serve static files. Replaced `socat TCP-LISTEN` with a direct bind. COOP/COEP headers for SharedArrayBuffer support.

[https://github.com/allen-munsch/rt-voice-wasm/blob/main/web/serve.jolt](https://github.com/allen-munsch/rt-voice-wasm/blob/main/web/serve.jolt)

### 2. Closed-loop voice scenario harness — `scripts/tts_harness.jolt`

Orchestrates edge-tts → ffmpeg → ws-send → jq event matching for five scenarios.

[https://github.com/allen-munsch/rt-voice-wasm/blob/main/scripts/tts_harness.jolt](https://github.com/allen-munsch/rt-voice-wasm/blob/main/scripts/tts_harness.jolt)

## Known limitations (potential future work)

- `getsockname` not bound — `getLocalPort()` on port 0 reports nothing
- `accept` blocks the calling thread — single-connection at a time

### DIFF
--- a/stdlib/jolt/socket.clj
+++ b/stdlib/jolt/socket.clj
@@ -1,0 +1,307 @@
+#!/usr/bin/env jolt
+;; java.net.Socket / ServerSocket / InetSocketAddress for Jolt via jolt.ffi.
+;; POSIX sockets + jolt.host/tagged-table + ref-put!/ref-get for state.
+;;
+;; Usage: (require 'jolt.socket)  ;; registers classes globally
+
+(ns jolt.socket
+  "POSIX socket support. Registers Socket, ServerSocket, InetSocketAddress."
+  (:require [jolt.ffi :as ffi]
+            [clojure.string :as str]))
+
+;; -- FFI --------------------------------------------------------------------
+(ffi/load-library)
+(def ^:private AF-INET 2)
+(def ^:private SOCK-STREAM 1)
+
+(ffi/defcfn c-socket     "socket"     [:int :int :int] :int)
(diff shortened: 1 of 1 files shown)


yogthos:
Merged, thanks! Follow-up in #543 hardened it: a write to a peer-closed socket now throws IOException instead of silently dropping (SIGPIPE guarded via MSG_NOSIGNAL/SO_NOSIGPIPE — without that a Linux server dies writing to a disconnected client), host strings passed to inet_addr/gethostbyname are NUL-terminated through string->ptr, and class/instance?/str answer as the mirrored classes.

One behavior change worth knowing for serve.jolt: ServerSocket(port) now binds the wildcard address like Java, not loopback-only. To keep a server local, pass a bindAddr: (java.net.ServerSocket. port 50 (java.net.InetAddress/getByName "127.0.0.1")). Also getLocalPort works with port 0 now (getsockname is bound), so the limitation in the description is gone.

PR #566, opened 2026-08-08
Fibers R7: the cheap park, chosen per park site

yogthos:
A go body that parks where the compiler can see it no longer captures a
continuation. The CPS pass in `clojure.core.async` rewrites the rest of the body
into a closure, and `__sm-take` / `__sm-put` store that closure and switch to the
scheduler with no `call/1cc` — so nothing holds a Chez stack segment while the
process is parked.

The choice is per **park site**, not per body, with the continuation park as a
runtime fallback. The round was originally specified the other way — prove every
park in a body is lexical, else compile the whole body to a continuation fiber —
which needs a closed-world call-graph analysis to answer "does anything this body
calls park", and a wrong answer there is a miscompile. That is what stopped the
first attempt. With a runtime fallback there is no static claim to defend: a park
the pass cannot see or cannot rewrite is left exactly as it is and parks the way
every park does today. No analyzer or backend change, no `eval`/`resolve` bail
rule, and a body can mix both mechanisms freely.

Measured on an M1 Max, both arms in one run, same process shape and retention so
only the mechanism differs (retained in a global, forced full collect, absolute
live bytes, peak RSS as the cross-check):

| parked process | live bytes | peak RSS |
| --- | --- | --- |
| rewritten body | 864 B | 1,244 B |
| captured continuation | 4,762 B | 7,368 B |

5.5x by live bytes, 5.9x by RSS. Not the 51x the spec predicted: that figure was a
bare closure, and a real process also carries the fiber record, its dynamic slice,
the waiter handler, the result channel and the resume closure. `mem-fiber-go`
re-derives at 4,123 B against R6's published 4,129 B, so the harness reproduces.

Park-and-resume round trip is 903 ns rewritten against 858 ns captured (three runs
each, spreads 38 and 75 ns) — inside the noise. It was 1.25x slower before two
fixes: the driver used `guard`, which costs a `call/cc` per entry and sat on the
resume path this exists to make cheap, and `jolt-fiber-resume*` guarded the thunk
on every entry, which for one of these resumes is redundant and also the wrong
handler (it marks the fiber dead without closing the body's channel, so a reader
would wait forever). The first harness ran both arms in one process and read the
second arm 25% slow whichever order they went in, so the arms are separate
processes now.

Coverage: `<!` `>!` `<!!` `>!!` reached through `do`, `let*`, `if`, `loop*`/`recur`
and application, which covers `let` `when` `when-let` `if-let` `cond` `and` `or`
`->` `doseq` `dotimes` and nested loops. A park inside a called function, a `try`,
a nested `fn`, a collection literal, an `alts!`, or reached through `eval` falls
back to the capture — each gated as a working capability rather than left to be
discovered. `alts!` and `try` coverage are follow-ups (jolt-nins, jolt-6jri).

Three bugs fixed in passing:

- `(>! ch nil)` on a fiber threw while **holding the channel mutex** and wedged
  that channel forever. The fiber put path holds the mutex by hand because it has
  to release before parking, so `with-mutex` is unavailable; the nil check moved
  out and the locked section now releases on any throw.
- `fibers-async.ss` defined `jolt-fiber-waiter-wait!` twice, and the live twin is
  the one `jolt-fiber-<!` depends on — reordering the file would have silently
  broken `<!`.
- `fibers-io-test.ss` did `string=?` on a `#f` result after a timeout, so an
  already-reported timeout crashed the gate and buried the failure.

Five later fixes to the pass itself, four of the same kind: it read a form
differently from the way the analyzer reads it.

`analyze-list*` accepts a `clojure.core` qualification on any head in its
`handled` set, because syntax-quote qualifies a macro like `letfn` whose
expansion the analyzer lowers as a special form. So `(clojure.core/let* ...)` and
`(let* ...)` are one form to the compiler and were two to this pass, and a
qualified head with a park inside fell to `sm-cps`'s `:else` arm and was rebuilt
as an application. That evaluated a `let*` binding vector, evaluated a `catch`
clause, quoted a gensym instead of the datum, and hoisted a park out of an `fn`
body, silently in each case. `sm-sf-head` strips the qualification and every
dispatch turns on that, which is what makes the `sm-opaque` comment's claim of
being true by construction actually true.

`:locals` starts from `&env` rather than empty. `sm-park-kind` was already right
without it, since `(resolve &env sym)` answers nil for a local, but `sm-expand`
calls the one-argument `macroexpand`, which has no env. So
`(let [or (fn [a b] :fn-called)] (go (or (<! c) :b)))` read `5` through the `or`
macro where the analyzer, which does shadow a macro with a local, gives
`:fn-called`.

`syntax-quote` was missing from `sm-opaque`. That set was written from
`clojure.core/special-syms`, but jolt's reader leaves a backtick as
`(syntax-quote datum)` for the analyzer where the JVM's reader expands it during
read, so `special-syms` never named it, and it was the one member of the
analyzer's `handled` that was neither rewritten nor opaque. It fell to the
`:else` arm and was rebuilt as an application, so `(go (list `(<! ch) :done))`
CPS'd the take inside the datum into a real park: the channel was drained and the
answer came back `(user/a__257 :done)` where the datum should have come back as
data. The set that has to be complemented is the analyzer's `handled`, not the
portable `special-syms`, and that is now asserted rather than restated.

`loop*` inits were collected with `take-nth` and handed to `sm-cps-seq` as a flat
argument list, which put every init in the scope OUTSIDE the loop.
`analyze-bindings` threads each `loop*` binding into the env the next one is
analyzed in, so `(loop [a 1 b (inc a)] ...)` is legal and `b`'s init means the `a`
above it. Flattened, that `a` resolved to whatever the enclosing scope happened to
call `a`: a compile error when nothing did, and silently the wrong value when
something did. `sm-cps-loop` binds sequentially now, the way `sm-cps-let` already
did, including when the init a later one depends on parks.

`jolt-fiber-resume*`'s sm arm was the one branch that could fall off the end
with no escape and no `jolt-fiber-done!`. Unreachable today, but a driver that
returned would strand the fiber in `'running` with its go channel never closed
and every reader of it waiting forever, with no error naming any of it.

The native `go` / `go-loop` in `async.ss` are gone. They expanded to a bare
`(go-spawn (fn* [] body...))` and the overlay's `defmacro` redefined the same two
vars, so which expansion a form got depended on whether the overlay had loaded,
and the native pair was dead wherever it mattered because the loader drops
`clojure.core.async` from `loaded-ns` precisely so a require pulls the overlay.
Both names stay reserved by a stub that says to require it, so the namespace is
not half-resolvable: `chan` answering bare while `go` reports "No such var" is a
worse failure than either.

Two fixes on the runtime side of the cheap park. `jolt-sm-fiber-spawn`'s thunk
opened with `(*txn* #f)`, copied from `jolt-fiber-go-spawn`, but that thunk is
re-entrant here: `jolt-fiber-resume*` runs it again on every cheap park, so the
line re-cleared `*txn*` on each resume right after `jolt-fiber-slice-restore!`
had put the fiber's own value back. Nothing reaches it today, since `dosync`
expands through an `fn*` that is opaque to the pass and a park inside a
transaction therefore takes the capture, and the line is unnecessary anyway:
`sa-fiber-spawn` builds the child's slice with txn `#f` by construction and
`jolt-fiber-run` restores that slice before the first entry. Removed rather than
moved, because the hazard is a once-per-fiber side effect living in a thunk whose
whole design is that it re-enters.

And the driver invariant is checked at the OP now, before the channel is touched.
Checked at the park it ran too late: the op had already appended a waiter to the
channel's list and committed the fiber to `'parked` under that waiter's `wmu`, so
the throw stranded a registered handler and a value delivered to it afterwards
went nowhere. The check stays at `jolt-sm-park!` too, unreachable now but that is
the point the invariant is load-bearing, and it is where the reasoning is written
down: the two field writes there are deliberately unordered, `sm` first leaves a
consumed continuation visible beside a pending step and `k` first leaves the step
missing while `sm` still reads `'running`, and `jolt-fiber-resume*` dispatches
wrongly on either. Neither order is safe alone. R0(d) is what makes it safe, since
the fiber is pinned to its carrier and only that carrier's thread dequeues it.

Gates: `test/chez/fibers-sm-test.ss` (53 checks, in `make fibers`) and
`host/chez/run-gosm.ss` (61 checks, new `make ci` target). Both were verified
non-vacuous by breaking each fix in turn: 9 checks fail with the pass forced off
while every value check still passes, which is the equivalence claim, 7 of the
qualified-head checks fail with `sm-sf-head` reduced to `sm-head`, the shadowing
check fails with `:locals` back to empty, 5 fail with `syntax-quote` out of
`sm-opaque` (4 on the datum plus the drift check, which reports the missing name
rather than an opaque failure), 2 fail with `sm-cps-loop` back to the flat
argument list, and scenario 12 leaves a waiter stranded on the channel with the
driver check back at the park.

The drift check is the one worth calling out, because it is what keeps
`sm-opaque` honest without anyone remembering to. It reads `jolt.analyzer/handled`
and `clojure.core.async/sm-opaque` as live vars and asserts that `handled` minus
the five heads the pass rewrites is a subset of `sm-opaque`, so a special form
added to the analyzer fails this gate on the day it lands instead of miscompiling
a go body later. `syntax-quote` is exactly what that would have caught. `make ci` green at 59
targets, self-host fixpoint holds, `make libconformance LIBS="core.async"`
unchanged (it flaps between `pass=210 fail=8` and `pass=211 fail=7` on this
branch and on its base, which is what the manifest's `:tolerance 2` is for).

---

### jolt-3907 (P1): shared runtime side-tables were unsynchronized

A Chez hashtable is not thread-safe, and unsynchronized mutation corrupts its
internals so the damage surfaces later inside the collector, naming nothing. Two
tables were fixed before this PR; auditing the rest found a worse one that the
bead never listed.

**`hasheq.ss`'s symbol and string caches.** They are written on every cache *miss*
from whatever thread is hashing, so every thread that puts a string or symbol in a
map writes one shared weak-eq table. 16 threads hashing fresh strings faults inside
the collector on **2 of 3 runs**; per-thread caches are clean 3 of 3. Deleting the
cache was not an option — recomputing is 8-19x (a 12-char key 47 ns vs 5.8 ns, a
48-char key 109 ns) — and neither was a mutex on that path. The vreg indirection
costs 5.86 → 7.36 ns on the bare hash (1.26x) and **158.5 → 162.5 ns on the path
that uses it** (1.025x), a string-keyed `pmap-get`.

Also: var interning is double-checked under a mutex (racing interns of one *new*
name could produce two cells for it, so a def through one would be invisible
through the other); var meta and the macro flag moved out of two eq-hashtables
**into the var cell**, which is safe and faster than the lookup it replaces;
`proc-name-tbl`'s check-then-set is one critical section; and the iref
watch/validator tables take a lock with a lock-free empty-table fast path — which
makes `def-var!` *cheaper* than before, since it used to probe both tables on
every def.

Verified already safe rather than assumed: `regex-cache`, `live-threads`,
`monitor-table`, `meta-table`, `variadic-fixed-arity-tbl` and
`thread-handles-by-id` all hold their locks on every path. Single-owner by
design: transients, the image walker, per-transaction tables. Written at load
and read-only after: `class-hint-table`, `class-token-alist`, `jhost-tag->fqn`,
`static-field-tbl`.

**The rest of the inventory is now done too**, on the same split: every mutation
and every whole-table scan takes the table's mutex, single-key reads stay
unlocked. That split is what makes it affordable, because the read paths here are
the hot ones. `find-protocol-method` is three probes on protocol dispatch,
`jch-tags` sits under it, and `chez-resolve-refer` runs per symbol per form in
the analyzer. None of them took a lock, and load and startup are unchanged at
0.45s and 0.11s.

**Multimethods.** `defmethod`, `prefer-method`, `remove-method` and the
`jolt-mm-epoch` bump were unlocked, so racing definitions dropped methods
outright. The dispatch cache is invalidated with `hashtable-clear!`, and a clear
can be undone: a resolve that began before a `defmethod` could publish its stale
answer after the clear, and every later dispatch would read it with the stamp
already saying current, so it never expired. Resolution reads the generation
before it walks and republishes only if that generation still holds. The
`.addMethod` and `.removeMethod` interop arms never bumped the epoch at all,
which was a plain correctness bug independent of threading.

**The protocol registry.** `register-protocol-method` builds `type-registry`'s
nested tables check-then-create, so two threads extending one type each made
their own inner table and an impl vanished with it.
`find-method-any-protocol` walked `hashtable-keys` unlocked, which under a
concurrent registration returns trailing FILL slots and then does a
`hashtable-ref` on `0` against a `#f` inner table, a crash on the dispatch path.
`intern-pm-key` minted two gensyms for one (proto . method) under a race, and
since the descriptor ptable is keyed by that identity, a store under one key and
a lookup under the other miss each other for the life of the process.

**The class graph.** The claim above that these hold their locks on every path
did not survive checking, which is why it now names different tables: only the
`hashtable-clear!`s were locked, not the cache writes and not `jvm-class-parents`
itself, whose read-modify-write dropped a concurrent graft. The two memo caches
get the same generation check as the multimethod cache. The worst one was
`jch-known-cache`, which published an empty table into the global and filled it
afterwards, so a concurrent `instance?` read it mid-fill and got a definitive
`#f` for a class the graph does model. It builds into a local and publishes once
complete.

Also locked: the `ns` alias / refer / refer-all / exclude tables and
`ns-registry`, `loaded-ns` and the AOT memos, `sys-prop-table`, the two FFI
tables, and `mutable-statics-tbl`.

Three new scenarios in `test/chez/thread-safety-test.ss`, each
checked against the unfixed code. Six and eight fail every run; seven loses a
protocol on 2 of 3, the rate scenario 1 reproduces at. Seven needed rewriting
once: the first version raced two protocols onto one shared tag and passed
unfixed, because the check-then-create window only exists on a tag's first
registration, which gave it two contended instants in 1200 iterations. Both
definers walk the same fresh tags in lockstep now.

**The inventory reopened, because this PR is what made it incomplete.** Loading
namespaces in parallel retires "written at load, read-only after" as a safety
argument, and the audit above was done before that landed. Five more tables, two
scenarios, `test/chez/thread-safety-test.ss` now 34 checks with all five new ones
failing against the unfixed code.

The serious one is **keyword interning** in `values.ss`, which the audit did not
name at all. Keyword equality IS identity, since `jolt=2-base` answers keywords
with `eq?`, and that is what makes `(:k m)` a pointer compare. These are strong
hashtables so nothing corrupts, but the intern was an unlocked check-then-set and
a lost update gives two `keyword-t` for one name. The hashes still agree, because
`khash` comes from ns/name, so a map lookup finds the right bucket, fails the
equality check, and answers nil for a key the map has. 8 threads over 4000 fresh
names split 64 of them, and `(get {:kw-0 42} :kw-0)` across a split came back
nil. Double-checked like `jolt-var` now, so the hot path stays the bare
`hashtable-ref` it was and only a first-ever intern takes the lock;
`symbol-string-pool` gets the same.

`jolt-register-callsite!` and `jolt-register-source!` are **emitted into every
compiled namespace** and run at load, so they run on several threads at once now,
and both were read-modify-write. 8 threads registering 6000 distinct sites each
dropped **297 of 48000** callsite entries. The source registry is worse than a
dropped entry: two threads registering one procname under different namespaces
both read `#f`, both wrote their own vector, and neither wrote `'ambiguous`, so
the trace names one of them with confidence. That is the misattribution the
marker exists to prevent, and it happened in **3 of 400** contended pairs.

`image-register-fn-form!` and `hc-stash-inline!` are the same shape, costing a
closure the image writer can no longer reconstruct and an inline that silently
does not happen.

`extensions.ss` had both failures the protocol registry and the multimethod epoch
had. The declare is check-then-create, so two threads declaring one point each
built their own providers table and a provider registered through one was
invisible through the other. And `bump-extension-epoch!` is a `set!`
read-modify-write, so a lost bump leaves a warmed call site serving the resolution
it cached before the new provider existed, with the stamp already saying current.
`register-tagged-methods!` had the identical check-then-create, and
`jolt-class-for` is another identity interner, milder only because the eq arm
compares `jclass` by name.

Every fix locks writes only; single-key reads of strong tables stay unlocked, on
the split `rt.ss` spells out for `var-table`, which is why none of it shows in the
timings.

Verified correct rather than assumed, so the inventory is closed: `records.ss`
under `rec-tbl-mu` throughout, `refs.ss`'s two weak history tables,
`regex-cache`, `jolt-register-variadic!`, the FFI tables, `jch-interface-set`, the
multimethod tables, `jolt-eval-marker-registry`. Boot-only and therefore
single-threaded: the `host-static.ss` registrars, `static-field-tbl`,
`embedded-resources`. Left alone on purpose: the two marker caches in
`source-registry.ss` are pure caches on the backtrace path with no identity or
correctness semantics, so a race there costs a recompute.

### jolt-psc5 (P2): a readiness registration could be lost

`jolt.io-poller` drained `:pending` in **two** critical sections — read, then
clear — so a registration landing in between was erased before it ever reached the
kqueue/epoll changelist. That fd's readiness was never reported and the fiber
waiting on it never resumed. Microseconds wide, which is why it appeared once as
one fiber of eight not finishing and never again in six isolated runs.

Same driver, 40 rounds of 8 fibers each parking on its own socket read:
**283/320 before, 320/320 after.** Not an edge-trigger issue (the poller is
level-triggered) and not in the commit/wake protocol, which was already correct.

New gates: `make threadsafety` (`test/chez/thread-safety-test.ss`, 7 checks, in
`make ci`) and `test/chez/poller-registration.clj` in `smoke.sh`. Both were
checked against the unfixed code: the first faults, the second reports
`LOST 4 of 8 readiness registrations in round 7`. The poller gate stops at the
first lost round, because unfixed it otherwise takes over eleven minutes to finish
what the fixed path does in under two seconds.

### Concurrent `require` was not serialized at all

Locking the loader's tables does not make `require` safe, so this went further.
Nothing serialized `load-namespace*`: two threads requiring one namespace both
passed the `loaded-ns` check and both ran its top-level forms, double-running
every `def` and side effect in the file. The mark-before-load that terminates a
require *cycle* made it worse across threads rather than better, because it marks
the namespace loaded before its forms run, so to a second thread it read as fully
loaded while still half-built and `require` returned having defined nothing.

The loader now follows JLS 12.4.2, the JVM's class-initialization procedure, per
namespace instead of per class. That is the design built for this shape:
initialization here is dynamic, re-entrant and allowed to be cyclic. Its step 3,
a recursive request by the same thread completing normally, is already jolt's
mark-before-load semantics, so nothing that works today changes; step 2, blocking
on another thread's in-flight load and re-checking when notified, is the part
that was missing. The load runs without the lock held, and a failure still rolls
the mark back so a retry can load.

Per namespace and not one global lock, so unrelated namespaces still load in
parallel and a slow load only blocks threads wanting that one. Clojure went the
other way, with `serialized-require` as `(locking RT/REQUIRE_LOCK (apply require
args))`, but it is private and its own docstring calls it an interim function,
and it serializes every load in the process. Go avoids the question entirely,
since package init is one goroutine, sequential, one package at a time, with
import cycles a compile error, and none of that is available here.

The JVM's own cost is that two threads entering a genuine cycle from opposite
ends deadlock, which OpenJDK closed as won't fix on JDK-8037567 and two earlier
reports. That part is not inherited: the owner of each in-flight load is already
recorded, so the wait-for graph is walked before blocking and we raise instead,
naming both threads and both namespaces. It changes nothing for a program that
works today. `:reload-all` over a require cycle also stops recursing forever,
since turning the dedup off for the whole extent had removed the mark as a
stopping condition and step 3 restores one.

New gate: `test/chez/concurrent-require.clj`, in `smoke.sh`. Eight threads gated
on a latch require one namespace; it asserts the target's top level ran exactly
once and that every thread saw the file's last form on return. Against the
unfixed loader it fails on every run, reporting 7 of 8 threads returning from
`require` before the namespace's last form had run. Also checked: a same-thread
`a`/`b` cycle still resolves both vars, `:reload-all` over a cycle terminates,
and two threads entering `x`/`y` from opposite ends report the cycle while one of
them still completes.

Ownership is by thread id, which is what makes step 3 a cycle break rather than a
self-deadlock. Two fibers on one carrier share a thread id, so the invariant is
thread-per-load rather than fiber-per-load. Nothing loading parks, but it is
stated at the definition rather than left to be discovered.

---

`make ci` green at 59 targets (`cli smoke` 126, `unit` 1224, corpus 4244/4263,
cts 6078 with the baseline 140 fail / 2 error, `gosm` 61, `fibers-sm-test` 53,
`threadsafety` 34). Self-host fixpoint holds. Load and startup unchanged at 0.45s
and 0.12s, measured on the release binary against this branch's base, including
against the branch without the locking commit so the added mutexes are accounted
for. `make libconformance LIBS="core.async"` unchanged.


yogthos:
Went back over this adversarially and pushed fixes for what it turned up. Filed the findings as beads first so the ones I did not close here have somewhere to live.

What I checked before changing anything: I ran about 40 go body shapes through the pass and diffed each against the same body spawned with `go-spawn`, which never gets rewritten. Side-effect ordering across a park, `case`, `and`/`or`, `when-let`, `cond`, destructuring, parks in recur arguments and in loop inits, closures capturing a post-park value, `binding` conveyance, `alts!` mixed with cheap parks in one body, nested `go`, exceptions raised inside a continuation. Every one matched. A 200 by 50 producer/consumer stress came out exact with 6512 cheap parks and zero captures. Compile time for the pass is linear, 3ms for 60 sequential parks and 3ms at 20 levels of nesting. The mechanism is solid and the fallback really does degrade the way the description says.

Three things in the R7 half were worth fixing.

The park counters were plain globals bumped with `set!` from whatever carrier thread parked. That is an unlocked read-modify-write, and `run-gosm` asserts exact deltas on them, so a lost increment is a flaky gate or a ten second timeout. They now live on the carrier record, where a carrier only touches its own field, and the two names became procedures that sum the pool. The bump stays lock free, which matters given where it sits.

`sm-opaque` was missing `case*`, `deftype*`, `reify*` and `import*`, and carried `syntax-quote`, which the analyzer never sees. None of the missing ones can reach the pass today, and I verified that by macroexpanding a `reify` and a `defrecord` inside a go body, both of which come out right because jolt's `case` expands to `let*` plus `if` and `reify` to a `make-reified` call. But the set is the thing that makes "a park the pass cannot rewrite is left exactly as it is" true, and it was true by accident of how those macros happen to expand rather than by construction. It is now `special-syms` minus the five heads the pass rewrites.

`jolt-fiber-resume*` set `'running` only on a first run, so a fiber resumed through either a continuation or an sm step ran as `'ready`. Nothing reads it, but the state machine at the top of fibers.ss and `alt-deliver!`'s race argument both name `'running` for a fiber that is executing. The set moved ahead of the cond.

Also made the guard in `jolt-chan-locked-give!` conditional on the channel having an xform. Both callers already hoist `async-check-put!` out of the mutex, so with the nil check done the only thrower left inside `ac-try-give!/locked` is the transducer step, and a plain channel no longer pays a `call/cc` per put on the path the rest of this round is shaving.

The jolt-3907 half had three gaps of the same shape, a mutex added where the fix was described but not everywhere the table is touched.

`proc-name-tbl` got a locked writer and a locked reader that nothing called. All five reads still did a bare `hashtable-ref`. Serializing writers against each other does not help when a reader is walking the table during a rehashing `set!`, and every one of those reads is cold, so they all go through `proc-name-of` now.

`iref-tbl-bump!` ran outside the mutex and after the table write, while the comment claimed the opposite of both. The ordering is the real problem: it left a window where a watch was in the table but `iref-any?` still read 0, so a def on another thread would silently skip validation and notification for a watch that was already registered. It is inside the mutex and ahead of the write now, and renamed to `iref-writes` because it counts calls and never decrements.

`var-table` had three writers taking no lock at all: `declare-var!`'s insert, the `ns-has-vars-set` write in `def-var!`, and `remove-ns`'s sweep, which also snapshotted `hashtable-keys` while another thread could be interning. All three take `var-table-mu` now and `declare-var!` double-checks like `jolt-var`.

I left the read side unlocked and changed the comment to say so plainly rather than implying the table is safe. Locking it measured 70 to 95 ns on a var-table probe and `var-deref` is emitted per var reference, 572 of them in the seed prelude alone, so closing it needs a design that keeps the read cheap, probably copy-on-write behind a box or a single-threaded fast path shared with the lazy-seq work. That is jolt-asj rather than something to guess at in this PR.

Last one, the poller gate kept only the output stream of each accepted socket, so 320 descriptors leaked over 40 rounds and the ServerSocket was never closed. Under a low `ulimit -n` that fails on accept for a reason unrelated to the bug it gates.

Beads: jolt-asj (P1, the var-table read side), jolt-73g, jolt-5dm, jolt-8k8, jolt-ifk, jolt-22b, jolt-cak, jolt-dsk, jolt-tiv. Everything except jolt-asj is closed by these commits.

`make ci` green at 59 targets twice, `make fibers` at 212 checks across seven gates, `make libconformance LIBS="core.async"` unchanged at `tests=31 pass=210 fail=8` (it reads 211/7 on some runs, which is the flap the manifest's `:tolerance 2` already documents, confirmed against an unmodified build).

yogthos:
Researched the var-table read side properly and it turned out I had the finding wrong, so this pushes a correction along with the fixes it did turn up.

I read Chez 10's hashtable source. There are two `adjust!` implementations and they behave oppositely.

A strong general hashtable, which `var-table` is (`s/newhash.ss:188`), never mutates existing structure. An insert is one `vector-set!` of a fully-formed list, an update is one `set-cdr!` of a value, and a resize builds a complete new bucket vector out of fresh spine conses, leaves the old one untouched, and publishes it with a single `ht-vec-set!`. `$gen-hashtable-ref` snapshots `ht-vec` once and takes its mask from that same vector, so the index cannot be torn. An unlocked single-key read can only ever observe a stale miss, and `jolt-var`'s double-check turns a stale miss into a lock acquisition and a re-read.

A weak eq hashtable (`s/library.ss:1570`) does the opposite: `($set-tlc-next! b ...)` relinks live cells in place while a reader may be walking them, and `lookup-keyval` is unsafe primitive code that stops only on a fixnum.

Verified both ways. 16 threads each probing then inserting 15k fresh keys, unlocked: the weak table hangs every run, the strong one survives every run. Then on the strong table, isolating just the writer lock with reads left unlocked throughout: unlocked writers lost 8,573 and 8,628 entries of 240,000 with no crash, serialized writers gave 240,000 of 240,000 twice. So the writer serialization from the last push is the complete fix for single-key access, and the lost entries are what justify it. A vanished cell means the next lookup interns a fresh one and the previous root is gone.

So the var-table read stays unlocked and the comment now explains why with the source cited, instead of calling it a known gap. Locking it costs 70 to 95 ns on a probe and buys nothing.

What did need fixing is the whole-table scans. `hashtable-keys`/`values` read `ht-size`, allocate a zero-filled vector of that length, then walk the buckets bounded by both, so a concurrent insert truncates and a concurrent `remove-ns` leaves trailing fill slots. Every caller immediately does `(var-cell-ns c)`, which on a fill 0 is an error that takes down whatever was running. 3,000 scans racing a grow/shrink writer produced 1.66M slots that were fill rather than entries. All six sites now snapshot through `var-table-cells` / `var-table-entries`, which hold the lock for the snapshot only so no caller's body runs under it.

The same split reprioritized what was left on jolt-3907. Sorting by table kind rather than by file, `ns.ss`, `loader.ss` and `multimethods.ss` are all strong tables, so they are the benign end. The urgent ones were two weak tables nobody had flagged: `rdr-map-order` and `rdr-ctor-forms` in `reader.ss`, and the deftype ctor-tag table in `records.ss`. Both are fixed, with different mechanisms because they have opposite profiles. The reader tables take a mutex, since they are written once per map literal read and copying per write would be quadratic in the size of the source. The ctor-tag table goes copy-on-write, since it is read on multimethod dispatch and written only at deftype definition, so a mutex would sit on dispatch to protect a write that happens a few hundred times in a process.

Both are gated and both gates were checked against the unfixed code. Pulling the reader mutex gives `nonrecoverable invalid memory reference`. Scenario 5 asserts the copy-on-write mechanism structurally rather than hammering it, because a rare-writer table cannot be made to fault on demand and the synthetic version passed unfixed at 200 ctors and again at 20,000; reverting the writer to mutating in place fails its first two checks with no timing at all.

One thing this turned up that is not a thread-safety issue and is not in this PR. A var reference at a site the per-site cell cache does not cover costs about 112 ns, against 28 ns at a cached site and 11 ns through a local. The cache only fires inside a `def`, because that is where `emit-with-cells` can hang the cell, so a bare top-level form resolves per access. An instrumented build confirms it: 3M calls through cached sites do 2 var-table reads, the same through uncached sites do 3,000,000. That is filed as jolt-g3u with the measurements and three candidate approaches. It is a compiler change with a self-host fixpoint to respect and should not ride along here.

`make ci` green at 59 targets. Load and startup unchanged against a build without these locks: 0.44s to require clojure.test, string, set and walk, 0.11s startup.

Beads: jolt-asj rewritten to the scan bug and closed, jolt-g3u opened for the var-deref cliff.

yogthos:
Took jolt-g3u, the var-reference cost I flagged last time. The fix is smaller than I expected and the interesting part is what I had to leave out.

The per-site cell that caches a resolved var, protocol impl or ctor descriptor only exists inside a def, because `emit-with-cells` is what creates the collector and only a def's init runs under it. `emit-top-form` now opens one around a top-level expression too, so a bare top-level `(loop …)` resolves each site once instead of once per iteration. That form goes from 120.8 to 25.1 ns/op, the same figure as the identical loop written inside a `defn`.

Two restrictions, and finding the second one is most of the work.

Not a def, because a direct-link def emits a bare `(define …)` that cannot sit inside a `let`, and both def paths already wrap their own init.

And only when the form contains a loop or a self-recurring fn, not merely a fn. That looks over-cautious and is not. A cell has to be paid for: the emitted `(var-cell-deref (or c …))` is several times the text of a `(var-deref …)`, and a top-level form is compiled by Chez every time it is loaded, so a cell only wins if its site actually runs again. A loop proves that. A top-level fn does not, since it may never be called. Including `:fn` cost 14% on `require clojure.test`, 0.44 to 0.50s, and bought that path nothing, because a namespace being loaded is defs and defmethods whose bodies run once at load.

Worth recording how that came out, since I got it wrong twice on the way. I first blamed the const pool that `emit-with-cells` also installs, on the theory that hoisting every keyword literal into a wrapper `let` was making large forms expensive for Chez to compile. A cells-only variant was equally slow, so that was wrong, and the cost is the size of the emitted cell form itself. I also nearly attributed the whole thing to build-to-build variance, so the final attribution was done by switching the wrapper at runtime inside one binary: same build, wrapper on 0.50s, wrapper off 0.44s.

The shape still paying full price is a closure handed to a higher-order fn from a top-level form, `(map (fn [x] (foo x)) coll)` at the REPL. Fixing that needs something better than a static guess about whether a fn repeats, and jolt-g3u carries the options and the measurements so nobody re-runs this experiment from scratch. Also still open there: `var-deref` builds the `"ns/name"` key with a fresh `string-append` on every probe, which is independent of the cell work and would help the compile path too.

Gated on `var-cache?`, which the mint runs with off, so mint output is unchanged and the self-host fixpoint holds. The tree walk that tests a form stops at a `:const` rather than descending into a literal that could be arbitrarily large.

One test needed updating, and I want to be clear it was a test problem and not a behaviour change. aot-cache-smoke's (o) case extracted which namespaces an artifact interns by matching `in-ns` and its argument as neighbours in the emitted text, and a cached var reference now separates them. It matches through a window instead, which is what the check actually wants and is independent of how a var ref is spelled. Its behavioural assertions passed throughout, and I confirmed the artifact still interns exactly `olib.top`.

`make ci` green at 59 targets, `make selfhost` reports the rebuild equals the checked-in seed, `make libconformance LIBS="core.async"` unchanged, load and startup flat at 0.44s and 0.11s.

PR #583, opened 2026-08-11
io-thread runs its body on a fiber

### LINKED ISSUE #579: implement core.async’s io-thread
All three can be available now without options. You can use (thread ...) for a 1:1 real OS thread. (go ...) for a stackless CPS transform and (io-thread ...) for a fiber.

yogthos:
Closes #579.

core.async names three carriers and jolt has all three now, with no option to set. `thread` is a real OS thread, `go` is the CPS pass on whatever `*go-backend*` says, and `io-thread` is a fiber. Before this, `io-thread` was an alias for `thread`, so asking for the cheap blocking-shaped carrier got you the expensive one, and `thread-call`'s workload argument was accepted and ignored.

The host gains `clojure.core.async/fiber-spawn`, which reaches `jolt-fiber-go-spawn` unconditionally. That is deliberately not a dispatcher: `thread-spawn` asks for an OS thread by name and gets one whatever `*go-backend*` says, so asking for a fiber by name has to work the same way. Otherwise `io-thread` would mean "a fiber, unless something above me bound the backend to `:thread`", which is not a thing a caller can use. `thread-call`'s workload now selects the carrier: `:io` is a fiber, `:mixed` (the default, as on the JVM) and `:compute` are OS threads, and anything else throws instead of quietly running on one of them. `:compute` stays a thread on purpose even though preemption keeps a compute-bound fiber from starving its carrier's queue, because fibers are capped at the carrier count, so N compute bodies on N carriers would leave nothing to run the io-thread bodies that are ready.

There is no CPS pass on an io-thread body. It parks by capturing its continuation, so a park works anywhere in it, including inside a called function, a try or a loop the pass could not rewrite. That is what a blocking-shaped body wants and it is the reason to have `io-thread` as well as `go`. What a fiber does not get is an OS thread of its own, so a body that blocks somewhere the runtime cannot see (`Thread/sleep`, a raw fd read, a blocking FFI call) pins its carrier for the duration. Channel ops park and so does `jolt.socket`, and `thread` is still the escape for everything else. The docstrings say which is which.

```clojure
(require '[clojure.core.async :as a])
(a/<!! (a/io-thread (jolt.host/fiber?)))   ;=> true
(a/<!! (a/thread    (jolt.host/fiber?)))   ;=> false
(a/<!! (a/thread-call (fn [] (jolt.host/fiber?)) :compute))  ;=> false
```

## Testing

`test/chez/async-io-thread-test.ss` is the gate, wired into `make fibers`. It runs with the pool pinned to one carrier, which is what gives its main check teeth: eight io-thread bodies parked at the same time on that single carrier, all of them resuming, and the park counter asserted to have moved by at least eight so "they all finished" cannot be satisfied by bodies that never parked. It also pins the carrier each name picks, that neither `io-thread` nor `thread` follows `*go-backend*`, the workload table, a park two frames deep in the body, a throwing body closing its channel and reporting to `go-monitor`, and what the body inherits from its spawner. Reverting the `:io` arm to `thread-spawn` fails four of its twenty checks.

Three `unit.edn` rows cover the same surface end to end through the production loader. `make ci` passes on this tree (62 targets). Also checked by hand outside the gate: a fiber spawning a fiber, `go` inside an io-thread body and an io-thread inside a `go` body, `thread` inside an io-thread body, 500 io-threads parked on one rendezvous channel, and `pipeline` (which uses `go-loop` and `thread`, untouched).

## Not in this PR

`go`'s default backend is unchanged, so "go is a stackless CPS transform" still needs `(binding [*go-backend* :fiber] ...)` to run on a fiber. Flipping that default is the R6 decision and changes every existing `go`, so it is worth its own change.

While writing the gate I found a pre-existing bug that io-thread makes easier to reach: a fiber's uncaught-throw trace can name frames from a sibling fiber that ran earlier on the same carrier, because the tail-call site lives in a per-thread virtual register that fibers on one carrier share. It reproduces on main with `go` and a `:fiber` binding. Filed as jolt-k5wm with a repro and a fix direction, not fixed here.

No CHANGELOG entry, since 0.7.2 went out today and the changelog is written per release. Happy to add a section if you would rather it land with the feature.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,10 @@ values:
 # bare monitor-enter/monitor-exit halves across a fiber switch. A monitor is the
 # one lock in the runtime that wraps user code, so neither half of locks.ss's
 # premise — short regions, never spanning a park — holds for it.
+# async-io-thread-test.ss is the io-thread gate (jolt-579): core.async's third
+# carrier. It runs with the pool pinned to ONE carrier, which is what makes "8
+# bodies parked at the same time, all of them resuming" mean that a fiber released
+# its carrier rather than held it.
 fibers:
 	@$(CHEZ) --script test/chez/fibers-test.ss
 	@$(CHEZ) --script test/chez/fibers-state-test.ss
@@ -245,6 +249,7 @@ fibers:
 	@$(CHEZ) --script test/chez/fibers-preempt-test.ss
 	@$(CHEZ) --script test/chez/fibers-lock-test.ss
 	@$(CHEZ) --script test/chez/fibers-monitor-test.ss
+	@$(CHEZ) --script test/chez/async-io-thread-test.ss
 
 # The dynamic-var binding stack (jolt-3bo): lookup cost against binding DEPTH and
 # against the number of vars in one frame, push/pop throughput, and the two
--- a/host/chez/java/async.ss
+++ b/host/chez/java/async.ss
@@ -601,15 +601,33 @@
           (if (eq? bv dyn-no-binding) (var-cell-root cell) bv))
         jolt-go-backend-thread)))
 
-;; The R4 dispatcher: go/go-loop/go-spawn go through here and honor the var;
-;; thread/thread-call keep a real OS thread regardless (the documented escape
-;; for blocking work — a fiber would pin its carrier). jolt-fiber-go-spawn is
-;; defined in fibers-async.ss, which loads after this file; the reference
+;; The R4 dispatcher: go/go-loop/go-spawn go through here and honor the var.
+;; Nothing else does — thread is always an OS thread and io-thread is always a
+;; fiber, because both name their carrier at the call site (see async-fiber-spawn
+;; below and the workload table in the overlay's thread-call). jolt-fiber-go-spawn
+;; is defined in fibers-async.ss, which loads after this file; the reference
 ;; resolves at call time, before any :fiber spawn can happen.
 (define (async-go-spawn thunk)
   (if (eq? (go-backend-current) jolt-go-backend-fiber)
       (jolt-fiber-go-spawn thunk)
       (async-go-spawn-thread thunk)))
+
+;; (fiber-spawn thunk) — ALWAYS a fiber, whatever *go-backend* says. This is the
+;; other half of the thread-spawn bargain: thread-spawn asks for an OS thread by
+;; name and gets one, so asking for a fiber by name has to work the same way, and
+;; io-thread (the overlay's thread-call :io) is what asks. A dispatcher would make
+;; io-thread mean "a fiber, unless someone above me bound *go-backend* :thread",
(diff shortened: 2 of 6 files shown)


yogthos:
Reviewed for correctness. I read the four source hunks, ran `make ci` on this tree (62 targets, pass), and then probed the surface the gate does not reach. Everything I tried behaved the way the docstrings say it does, so this is good to merge.

The design call at the center of it is right. `fiber-spawn` has to bypass the dispatcher for the same reason `thread-spawn` does: a name that means "a fiber unless an enclosing `binding` disagrees" is not something a caller can build on, and the gate pins that in both directions (2c, 2d). The workload table matches the JVM, including `:mixed` as the default and a throw rather than a silent fallback for anything else. `thread` still expands straight to `thread-spawn`, so nothing that was an OS thread stopped being one.

Probes I ran outside the gate, all with the pool pinned to one carrier so that a body holding its carrier shows up as a hang rather than as a pass:

- a park on a `timeout` channel, on `alts!!` where the timeout wins, and on a channel that main later `close!`s (nil comes back in each case)
- an io-thread inside an io-thread, an io-thread inside a `:fiber` `go`, a `go` taking from an io-thread channel, and a `thread` inside an io-thread body
- 1000 io-thread bodies parked on one rendezvous channel, all 1000 resuming with the right value, plus 50 `thread-call ... :io` under contention
- `locking` around a park with two bodies contending for the same monitor, `dosync` in a body, a park inside a `dosync` followed by a second clean `dosync`, and 100 fiber transactions on one ref (the txn does not leak into the next one)
- `a/into`, `onto-chan!!` and `pipeline` over channels fed from io-thread bodies
- `put!`/`poll!`/`offer!` inside a body, a nil body versus a false body, binding conveyance through two levels of io-thread, and a throw several frames deep

One thing to flag, and it is not from this PR. The shared timeout timer in `host/chez/java/async.ss` has two bugs, both reproducible on main with no fiber anywhere:

```clojure
(a/<!! (a/timeout 100))          ;; 104 ms
(def far (a/timeout 3000))
(a/<!! (a/timeout 100))          ;; 2998 ms
```

`timeout-thread` sleeps to the nearest deadline with `timeout-mu` released, so it is not on `timeout-cv` when a nearer deadline is inserted. The signal is dropped and the new timeout fires at the old deadline instead of its own, which means any short timeout created while a long one is pending is late by the difference. Separately, when the pending list drains empty the timer clears `timeout-running?` before waiting, so the next insert both signals the waiting thread and forks another one: 100 sequential `(timeout 1)` calls leave 100 immortal timer threads (measured with `ps -M`, against 1 thread at baseline).

Neither is reachable through anything this PR added, and io-thread only makes the first one easier to notice, since a parked fiber resuming late looks like the carrier being held. I am merging this as is and fixing the timer separately.

PR #609, opened 2026-08-14
Profile standalone binary startup stages

sundbp:
Built executables now expose an opt-in JOLT_STARTUP_PROFILE trace across the native launcher, runtime files, application namespaces, and -main. Each Scheme marker reports interval wall and CPU time, collection count, reclaimed bytes, current heap size, and cumulative startup time to stderr; the normal path stays silent.

Gate both enabled marker coverage and unchanged disabled output in the standalone build smoke, and document how to profile any built executable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,17 @@ out — keeping everything, and reporting which library is responsible — when 
 code resolves vars by name at runtime (`eval`/`resolve`/`ns-resolve`/…). See
 [deps.edn internals](https://jolt-lang.github.io/docs/tools-deps.html) and [RFC 0007](https://jolt-lang.github.io/docs/rfc/0007-compilation-modes-and-binary-output.html).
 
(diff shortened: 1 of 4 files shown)


yogthos:
Merged — thanks, the instrumentation is what turned the ctk investigation actionable. Two things worth knowing on your side: current main compiles your two Link-header patterns in ~1ms (the SRE-direct translator landed a while back), and #610 goes further — regex engines now build lazily at first match, so clj-http.lite.links namespace init drops to parse-only (~23µs/pattern) and the wrap-links middleware costs nothing unless Link headers are actually parsed. A rebase + rebuild should erase the 372ms row and most of the finalization collection behind it; your profiler is the right tool to confirm the new numbers. Also note the launcher-stage markers ride the prebuilt stub, so they appear once your jolt itself is rebuilt at or after this commit.

PR #635, opened 2026-08-16
Resolve a chunk's leaf once per block, not once per element

yogthos:
First increment of epic jolt-r8tz (seq tier).

## What was wrong

jolt's map/filter/remove **already** consume chunks — `map-seq` has an
`na-chunked-seq?` branch, and realization on `(first (map f a-100-vector))` is 32
on both runtimes. The chunking simply wasn't paying for itself: `cseq-chunked`
read every element through `pvec-nth-d`, and `vec->seq` re-descended the trie per
element.

Over a 100k vector that cost **107.4 ns/elem against 79.7** for the same map over
a **list** — chunking was worse than not chunking.

## Fix

A standalone chunk is ≤32 elements and therefore entirely tail-resident, and the
leaf holding element *i* also holds the next 31. Resolve it once per block and
index it directly. `pvec-nth-d` measures 13.1ns against a raw `vector-ref`'s
1.9ns (Chez 10.4.1).

| 100k elements | before | after |
|---|---|---|
| `map inc` over a vector | 107.4 ns/elem | **61.9** (1.73x) |
| `count` of a vector seq | 59.5 | **30.8** (1.93x) |
| `map inc` over a list (untouched) | 79.7 | 81.1 |

The vector row now sits below the list row, which is the property that was
inverted.

## Gate

New `seqscaling`, in CI-GATES. Shape, not an absolute ceiling: per-element cost
over a chunkable source must beat the same over a list.

Two things the gate's design had to get right, both learned the hard way here:

- **n=100000, not 32.** At 32 the whole vector lives in its tail, so there is no
  descent to skip and the defect is invisible — 93.8 ns/elem before *and* after.
  My first version of this gate used 32 and was vacuous.
- **`count`, not `doall`.** `doall` retains n cells per repetition, so the two
  arms get compared under different GC pressure and the ratio wanders.

Red-proofed by reverting only `seq.ss` (with a forced rebuild — `make testbin`
reported "up to date" and quietly measured the fixed binary the first time):
the map section goes above its ceiling. The filter section passes either way
(0.84 → 0.68) and is documented as a coarse still-chunking check, not the sharp
one.

## Correctness

Cell shape and fields unchanged, so chunk-first / chunk-rest / chunked-seq? /
reduce see what they saw before. Verified across trie depths (32 / 100 / 5000
elements), subvec, chunk boundaries, and reverse.

Gates: unit, corpus (0 new divergences), values, selfhost, smoke, shakesmoke,
portcheck, chunkscaling, readscaling, vecscaling, seqscaling.

Still ~3.7x off the JVM on this path (61.9 vs 16.7 at 32 elements); the residual
is bead jolt-r8tz.3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ endif
 JOLT-TARGETS-NEEDING-DEPS := \
   aotcacheperf aotcachesmoke aotfingerprint asynctimer buildlibsmoke buildsmoke \
   aotcachepathsmoke compilepathsmoke contagion corpus cts dcerefs depssmoke depsunit devboot \
-  readscaling vecscaling pipescaling chunkscaling printscaling ioscaling hotscaling \
+  readscaling vecscaling pipescaling chunkscaling printscaling ioscaling hotscaling seqscaling \
   devbootsmoke devirt directlink ffi fibers fieldjoin fieldnum fieldread flarr fnform grenadine \
   gateboot gatebootsmoke gosm httpsfetch infer inline inline-body irvalidate \
   jolt jolt-debug jolt-release joltsmoke libconformance mandelbrot-num mathfl mvnhttp \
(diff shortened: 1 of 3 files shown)


yogthos:
Closing: the claimed speedup does not exist.

Re-measured with forced rebuilds on both arms (make testbin silently reports 'up to date' after a git stash), same method, median of 5, 100k elements:

|                     | with fix | no fix |
|---------------------|----------|--------|
| doall map 100k-vec  | 86.2     | 87.4   |
| count map 100k-vec  | 52.9     | 52.7   |
| doall map 100k-list | 86.4     | 87.0   |
| count map 100k-list | 63.9     | 62.6   |

All within noise. The 107.4 -> 61.9 figure in the description was not a valid A/B — both arms ran the same (fixed) binary because the restore step had already executed, so it measured run-to-run variance.

Why the premise was wrong: a standalone chunk is at most 32 elements and therefore always tail-resident, so pvec-nth-d never descends the trie there — the branch it takes is the cheap one already. The 13.1ns I measured for pvec-nth-d in isolation does not reproduce in context, where Chez inlines it against a tail-resident chunk.

The gate went with it: vector-vs-list ratios gave 0.79, 1.07 and 1.25 for identical code depending on method, and the two CI jobs on this PR disagreed with each other. A depth-independence variant did not discriminate either (1.25 unfixed vs 1.33 fixed).

The underlying question — why chunking pays on the JVM and not here — is real and stays open on jolt-r8tz.

PR #642, opened 2026-08-17
Fiber-parking for jolt.process subprocess pipe reads/writes

### LINKED ISSUE #641: Fiber-parking for jolt.process subprocess pipe reads/writes
## Problem

`jolt.process`'s pipe I/O (`host/chez/java/process.ss`'s `proc-fd-input-port`/
`proc-fd-output-port`) does a blocking retry loop with zero `jolt.io-poller`
wiring. On the `:fiber` `core.async` backend, a blocked pipe read/write pins
the whole fiber carrier instead of parking — with more concurrent pipe-I/O
`go`-blocks than `*fiber-carrier-count*` (default = CPU count), the excess
ones silently starve forever.

## Why it matters

Fibers R8 already solved this exact problem for sockets — a socket read
parks the fiber instead of pinning the carrier — and R8's own changelog
explicitly deferred "file offload" as a separate round. Subprocess pipes
never got it. Any project driving concurrent subprocess I/O from fibers
(LSP-server brokers, build-tool wrappers, shell pipelines) hits this gap;
we hit it directly building an LSP-broker library on Jolt.

## Proposed fix

Mirror R8's own `jolt.socket` pattern onto `jolt.process`, entirely at the
Scheme/FFI layer:

- Non-blocking pipe fds at creation — role-aware (only the end the parent
  retains, since `O_NONBLOCK` is shared via `dup2`'s open-file-description;
  a role-blind version would make the *child's* own stdio non-blocking too).
- On EAGAIN, park via a small bridge into `jolt.io-poller/wait-ready`, which
  is already fd-generic — **zero changes needed to `io_poller.clj` itself**.
- A real fallback (not a busy-loop) when no poller is loaded, so
  `jolt.process` used standalone pays zero behavior change.
- `forget!` the fd from the poller on close, matching `socket.clj`'s own
  close-then-forget order (closes a use-after-free/stranding class of bug
  the same shape as one already fixed+changelogged for sockets).

No changes to `jolt.io-poller`, `jolt.socket`, or `jolt.process`'s public
Clojure API. Not general file-descriptor/file-offload support — scoped to
process pipes, the concrete need.

## Status

Implemented and verified: 6 commits, a permanent gate
(`fibers-process-io-test.ss`, 36 checks) wired into `make fibers`, all 12
existing `fibers` gates green, `process-test.clj` and the `process-suite`
conformance harness both confirmed unaffected. Opening a PR against this
issue.

burinc:
Closes #641.

## Summary

Extends Fibers R8 ("a socket read parks the fiber instead of pinning its
carrier") to `jolt.process` subprocess pipe reads and writes. Today a
blocked pipe read/write pins the whole fiber carrier instead of parking;
with more concurrent pipe-I/O `go`-blocks than `*fiber-carrier-count*`, the
excess ones silently starve forever. See #641 for the full problem
statement and rationale.

## What changed

Entirely at the Scheme/FFI layer in `host/chez/java/process.ss` — mirrors
the existing `jolt.socket` pattern, no changes to `jolt.io-poller`,
`jolt.socket`, or `jolt.process`'s public Clojure API:

- Non-blocking pipe fds at creation, role-aware (only the end the parent
  retains — `O_NONBLOCK` is shared via `dup2`'s open-file-description, so a
  role-blind version would make the *child's* own stdio non-blocking too).
- A working `fcntl` binding for Apple arm64: `F_SETFL` is C-variadic, and
  without the `(__varargs_after 2)` calling-convention marker it reports
  success but silently never applies the flags.
- On `EAGAIN`, park via a small bridge into the already fd-generic
  `jolt.io-poller/wait-ready`; a real blocking fallback (not a busy-loop)
  when no poller is loaded, so `jolt.process` used standalone pays zero
  behavior change.
- `forget!` the fd from the poller on close, matching `jolt.socket`'s own
  close-then-forget order — otherwise a close racing a parked read/write
  strands the fiber permanently, and a leaked tombstone in the poller's
  shared fd-keyed table can be consumed by a later, unrelated socket
  assigned the same reused fd number.
- A per-port `closed?` flag, set synchronously by the close proc before it
  frees/closes/forgets, checked at the top of every retry — the
  `forget!` wakeup is asynchronous, so without this a woken fiber's retry
  could execute a real syscall against an already-freed buffer and a
  possibly-reused fd.

## Testing

New permanent gate, `test/chez/fibers-process-io-test.ss` (36 checks),
wired into `make fibers` right after the socket half's own gate. Covers:
real parking under contention (read and write sides, each with a sibling
fiber proven to make progress while the first stays parked), thread-mode
(off-fiber) behavior, a full GC succeeding mid poller-wait, N fibers on one
carrier each reading a separate subprocess pipe, the existing
`jolt.process` API unaffected off-fiber, the no-poller fallback, and two
close-safety regressions (a close racing a parked read must not strand the
fiber; a close racing a parked write must take the closed short-circuit,
not re-run the syscall against freed memory — this one is a deterministic
mechanism-proof test, not dependent on winning a real fd-reuse timing race).

`make fibers`: all 12 gates green. `bin/jolt run test/chez/process-test.clj`
and the `process-suite` conformance harness both confirmed unaffected.

## Scope

Process pipes only, not general file-descriptor/file-offload support — R8's
own changelog deferred that explicitly, and it's out of scope here too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Fibers R8, extended to `jolt.process`: a subprocess pipe read or write
+  parks the fiber instead of pinning its carrier.** The trick `jolt.socket`
+  already applies to sockets now applies to `proc-fd-input-port` /
+  `proc-fd-output-port` (`host/chez/java/process.ss`): the pipe fd the
+  parent retains is set `O_NONBLOCK` at creation (never poller-gated), and
+  on `EAGAIN` registers readiness with `jolt.io-poller` and parks the
+  current fiber — the carrier runs other fibers meanwhile, and blocks on a
+  private `kevent`/`epoll_wait` exactly as before when there is no fiber
+  (jolt.process's own no-poller fallback keeps working standalone). A
+  working `fcntl` binding for Apple arm64 is what makes the non-blocking fd
+  viable at all: `F_SETFL` is C-variadic, and without the
+  `(__varargs_after 2)` calling-convention marker it reports success but
+  silently never applies the flags. Closing a pipe port now also tells
+  `jolt.io-poller` to forget the fd, mirroring the socket fix below: a
+  closed fd is auto-dropped from the kernel's kqueue/epoll set, so a fiber
+  still parked on it would otherwise wait forever for an event that is
+  never coming. Gated by `test/chez/fibers-process-io-test.ss`, wired into
+  `make fibers` right after the socket half's own gate.
+
 ## [0.7.14] - 2026-08-16
 
 A collections release. The theme running through it is operations that were
--- a/Makefile
+++ b/Makefile
@@ -247,6 +247,7 @@ fibers:
 	@$(CHEZ) --script test/chez/fibers-go-test.ss
 	@$(CHEZ) --script test/chez/fibers-pool-test.ss
 	@$(CHEZ) --script test/chez/fibers-io-test.ss
+	@$(CHEZ) --script test/chez/fibers-process-io-test.ss
 	@$(CHEZ) --script test/chez/fibers-sm-test.ss
 	@$(CHEZ) --script test/chez/fibers-preempt-test.ss
 	@$(CHEZ) --script test/chez/fibers-lock-test.ss
--- a/host/chez/java/process.ss
+++ b/host/chez/java/process.ss
@@ -413,9 +413,23 @@
   (for-each sa-foreign-free (cdr m))
   (sa-foreign-free (car m)))
 
-;; binary ports over the raw pipe fds. EINTR is retried; any other read error
-;; reads as EOF (the child is gone and the pipe with it). A short write returns
-;; its count and Chez's port machinery re-calls for the rest.
+;; binary ports over the raw pipe fds. EINTR is retried; EAGAIN parks the
+;; current fiber (or blocks this thread) via jolt.io-poller when it's loaded
+;; -- symmetric on both sides below: proc-fd-input-port waits on :read,
+;; proc-fd-output-port waits on :write, over the same mk-pipe-created
+;; non-blocking fd and the same poller-wait-ready bridge. When no poller is
+;; loaded, EAGAIN clears O_NONBLOCK on this fd (fd is genuinely non-blocking
+;; now per mk-pipe above, so a no-poller EAGAIN is a live "not ready yet"
+;; signal, not a real failure) and falls through to a real blocking
+;; proc-c-read/proc-c-write from then on -- exact behavioral parity with
+;; pre-this-plan code once triggered, verified standalone: a blocked
+;; read/write genuinely waits rather than busy-looping or returning a
+;; spurious result. Any OTHER error still ends the port, but the two sides
+;; report it differently -- pre-existing asymmetry, unchanged by this: the
+;; read side reads it as EOF (0, the child is gone and the pipe with it);
+;; the write side raises (child gone, `error 'process "write to child
+;; failed"`). A short write returns its count and Chez's port machinery
+;; re-calls for the rest.
 (define proc-fd-buf-size 32768)
 (define (proc-fd-input-port fd)
   (let ((buf (sa-foreign-alloc proc-fd-buf-size)))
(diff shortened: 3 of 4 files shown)


yogthos:
Oh nice, I was going to come back to figure out how to move carriers. Glad you got it figured out! :)

yogthos (approved):
Reviewed the diff and ran `make fibers` locally: all 12 gates green, the new `fibers-process-io-test.ss` reports 36 checks / 0 failures.

The core of this is right, and the two subtle parts are the parts I'd have expected to get wrong. Role-aware `O_NONBLOCK` is correct and not obvious: `dup2` shares file status flags through the open file description, so flipping the end that goes to the child really would give the child spurious `EAGAIN` on its own stdio, and the `#f #t` / `#t #f` per-role split is exactly the right shape. The `closed?` flag's ordering argument also holds up: `set-box!` before `poller-forget!`, and `forget!` reaches `sa-fiber-resume`, which takes the carrier mutex that the carrier itself must acquire in `jolt-fiber-dequeue!` before the woken fiber can run, so the `#t` is genuinely published. Close-then-forget matches `socket-close!`. The test file is thorough, and check 9 asserting *which* path the woken fiber takes rather than just "nothing crashed" is the right instinct.

One finding is significant enough that I want to flag it clearly, then three smaller ones.

## The fix does not engage in the case #641 describes

`poller-wait-ready` only uses the poller if `jolt.io-poller` has already been loaded by someone else, and the only thing in the tree that requires it is `jolt.socket`:

```
$ grep -rn "io-poller" --include="*.clj" --include="*.ss" . | grep -v test/
host/chez/locks.ss:189:          ;; (comment)
host/chez/java/concurrency.ss:828:  ;; (comment)
stdlib/clojure/core/async.clj:507:  ;; (comment)
stdlib/jolt/socket.clj:15:       [jolt.io-poller :as poller]
```

So a program that uses `clojure.core.async` fibers plus `jolt.process` and never touches a socket — a build-tool wrapper or shell pipeline, two of the three cases the issue names — finds `wait-ready` unbound, takes the no-poller fallback, clears `O_NONBLOCK`, and blocks. The gate does not catch this because checks 1-5 require `jolt.io-poller` explicitly, and check 6's probe requires only `jolt.process`, so neither ever exercises "fibers are in use but nobody loaded the poller".

I confirmed it against the branch. Requiring `jolt.process` and `clojure.core.async`, one carrier, a fiber reading a subprocess that sleeps 3s, and a sibling fiber spawned after it:

```
jolt.io-poller/wait-ready bound? #f
same carrier? #t
reader state while sibling waited: running
sibling finished within 1.5s (child sleeps 3s)? #f
VERDICT: STARVED — carrier pinned by the pipe read
```

That is #641 verbatim, on the branch that closes it. The machinery here is all correct; what is missing is making the poller reachable. The right shape looks like a lazy `load-namespace "jolt.io-poller"` in `poller-wait-ready`, latched once and taken only when `(jolt-current-fiber)` is non-#f, so a plain thread keeps the cheap blocking fallback and never pays a private kqueue per wait. The loader's claim protocol already handles concurrent and fiber-parking loads (`ldr-begin-load!` steps 2-3), so that call is safe from inside the port read. I'll take this as a follow-up rather than hold the PR, since everything else here is a prerequisite for it.

## Widening `proc-spawn-fd-ok?` demotes too much

Adding `proc-fcntl-get proc-fcntl-set proc-errno-loc` to that `and` means a machine where any of the three fails to resolve loses the entire `posix_spawn` path and falls back to Chez's `fork`. The comment 20 lines above explains why that fallback is the fallback: it leaves `SIGINT` at `SIG_IGN` in the child, and it emulates INHERIT. Trading correct child signal dispositions for the ability to set `O_NONBLOCK` is the wrong way round. Better to leave the gate as it was and make `set-nonblocking!` a no-op when the bindings are absent: the pipes stay blocking, `EAGAIN` never arrives, and the only thing lost is parking. The `proc-errno-loc` half is the same story, just with a sharper failure mode (no errno means `EAGAIN` reads as EOF), so the answer is still "don't set `O_NONBLOCK`", not "abandon `posix_spawn`".

In practice `fcntl` resolves everywhere this runs, so this is latent rather than live.

## `set-nonblocking!` / `poller-wait-ready` / `poller-forget!` should carry the `proc-` prefix

The rationale given is that these are bridge points reachable beyond the file, but nothing outside `process.ss` references any of the three, and the file's own header (around line 40) documents why an unprefixed top-level define here is a hazard: this file is `load`ed at top level, so a second define of the same name silently wins. `proc-libc-signal` exists precisely because of that. These three are private helpers and reintroduce the exposure the header warns about.

## Two small ones

The test file's watchdog comment says "the whole workload (all 6 checks)" but there are nine now.

`stdlib/clojure/core/async.clj:507` lists what parks on a fiber and what pins the carrier, and singles out `jolt.socket`. Once this lands, `jolt.process` pipe I/O belongs in that sentence, and "a blocking read on a raw fd" no longer covers subprocess pipes.

## On comment volume

The 60-line block before `proc-fd-input-port` and the 20-line one inside `mk-pipe` are heavy even by this file's standards, and some of it restates the code immediately below. The reasoning is good and I would keep the ordering argument and the `dup2` file-status-flags explanation; the parts I would cut are the ones enumerating what is deliberately not covered, which read more like PR description than like something a future reader of this function needs.

Merging. Follow-ups for the poller reachability, the `and` gate, the prefixes, and the two doc lines.

PR #643, opened 2026-08-17
Make the subprocess pipe fiber-parking actually engage (jolt-641)

yogthos:
Follow-up to #642, which extended Fibers R8 to `jolt.process` subprocess pipes. The machinery there is right, but it does not engage in the case #641 describes, and this finishes it. Also carries two smaller cleanups from reviewing that PR.

## The poller was not reachable

`proc-poller-wait-ready` only used `jolt.io-poller` if something else had already loaded it, and `jolt.socket` is the only namespace in the tree that requires it. So a program that drives subprocesses from fibers and never opens a socket, which is two of the three cases #641 names (a build-tool wrapper, a shell pipeline), found `wait-ready` unbound, took the blocking fallback, cleared `O_NONBLOCK`, and starved exactly as before.

Verified on the merged branch, two carriers with both hosting a subprocess pipe read and a third fiber body spawned after them:

```
poller loaded by this program's requires? false
sibling result: nil after 1509 ms
VERDICT: STARVED — every carrier pinned by a pipe read
```

and on this branch, same program:

```
poller loaded by this program's requires? false
sibling result: :sibling-ran after 1 ms
VERDICT: the pipe reads parked — a further body ran on a full carrier pool
```

`proc-poller-wait-ready` now loads the poller itself, at the one moment it is known to be needed and only when there is a current fiber to park. A plain thread keeps the blocking fallback rather than paying a private `kqueue` per wait, which is the cheaper answer for it anyway and leaves every existing off-fiber `jolt.process` caller on exactly the path it was on.

The load leans on `loader.ss`'s claim protocol rather than on a latch of its own, and that turned out to be the part that matters. Two fibers on two carriers reach their first `EAGAIN` at once routinely, and my first attempt used a boolean "already tried" latch set before the load. The second fiber skipped straight past it, found the var still unbound because the first was only part way through, and cleared `O_NONBLOCK` on its own fd for good, so that port never parked again. One of the two readers sat `:running` while the other sat `:parked`. Calling `load-namespace` instead makes the second fiber wait on the first's claim, fiber-aware so it parks rather than pinning, and then find a fully built namespace. A load that fails is recorded and not retried, since the loader deliberately rolls its own mark back and nothing about the next `EAGAIN` will have changed.

Narrowing `proc-spawn-fd-level`'s lock falls out of this. The port constructors can now reach a park and were being called inside the `proc-spawn-fd-mutex` region, which `make parkcheck` refuses outright. Only their `read!`/`write!` callbacks can actually park and those run when someone reads the port, so nothing was wrong, but the lock's own contract is `pipe()` through `posix_spawn` and nothing else. It hands back raw fds now and the ports are built after it returns.

## The gate check that should have caught it

Check 6 required `jolt.io-poller` explicitly for checks 1-5, and its standalone probe only asserted that a read eventually returned the right bytes, which the fallback also does. It now covers both halves of the choice, in the one order that can observe them: off a fiber the poller must stay unloaded and the read must really block, then on a fiber it must autoload and both readers must park with a third body proving the carriers are free. Two carriers and two readers rather than one of each, because the concurrent autoload is the case with a wrong answer available.

Both failure modes were confirmed to fail the check. Without the autoload at all: `parked2=#f sib-ran=#f post=#f`. With the latch instead of the claim protocol: `parked2=#f sib-ran=#f post=#t states=(done done)`.

## Two cleanups

`proc-spawn-fd-ok?` had grown `proc-fcntl-get`, `proc-fcntl-set` and `proc-errno-loc`, so a machine where any of the three fails to resolve stops using `posix_spawn` altogether and falls back to Chez's fork. The comment above `proc-spawn-fd-mutex` explains why that is the fallback and not the default: it leaves `SIGINT` at `SIG_IGN` in the child. Correct child signal dispositions are not worth trading for the ability to set `O_NONBLOCK`, so the parking bindings get their own gate and the degradation when they are missing is that pipes stay blocking.

`set-nonblocking!`, `poller-wait-ready` and `poller-forget!` were left unprefixed as bridge points reachable beyond the file, but nothing outside `process.ss` references any of them, and the file's own header documents why an unprefixed top-level define here is a hazard: the file is `load`-ed at top level, so a second define of the same name silently wins. That is why `proc-libc-signal` exists. They carry the prefix now.

## Testing

`make fibers` all 12 gates green, `fibers-process-io` 36 checks / 0 failures. `make parkcheck`, `make lockcheck`, `make gosm`, `make asynctimer` green. `bin/jolt run test/chez/process-test.clj` and `host/chez/process-suite.sh` (the vendored babashka.process upstream suite) both clean. `make test` in progress; I will note the result here.

The CHANGELOG entry for #642 is still unreleased, so it is amended in place rather than given a second bullet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   on `EAGAIN` registers readiness with `jolt.io-poller` and parks the
   current fiber — the carrier runs other fibers meanwhile, and blocks on a
   private `kevent`/`epoll_wait` exactly as before when there is no fiber
-  (jolt.process's own no-poller fallback keeps working standalone). A
-  working `fcntl` binding for Apple arm64 is what makes the non-blocking fd
-  viable at all: `F_SETFL` is C-variadic, and without the
-  `(__varargs_after 2)` calling-convention marker it reports success but
-  silently never applies the flags. Closing a pipe port now also tells
+  (jolt.process's own no-poller fallback keeps working standalone). A fiber
+  that is about to park autoloads `jolt.io-poller` if nothing else has
+  required it, which is what makes this reach the programs that motivated
+  it: `jolt.socket` was the only namespace in the tree that pulls the
+  poller in, so a build-tool wrapper or shell pipeline driving subprocesses
+  from `go` blocks without ever opening a socket would otherwise have found
+  the poller absent and taken the blocking fallback — pinning its carrier,
+  the exact starvation this removes. A plain thread does not autoload it and
+  keeps the cheaper blocking read. A working `fcntl` binding for Apple arm64
+  is what makes the non-blocking fd viable at all: `F_SETFL` is C-variadic,
+  and without the `(__varargs_after 2)` calling-convention marker it reports
+  success but silently never applies the flags. Missing `fcntl` or `errno`
+  bindings cost only the parking, not the `posix_spawn` path itself, whose
+  fallback to Chez's `fork` leaves `SIGINT` ignored in the child. Closing a
+  pipe port now also tells
   `jolt.io-poller` to forget the fd, mirroring the socket fix below: a
   closed fd is auto-dropped from the kernel's kqueue/epoll set, so a fiber
   still parked on it would otherwise wait forever for an event that is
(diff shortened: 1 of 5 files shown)


yogthos:
`make test` is green: `OK: ci gate passed (79 targets)` / `OK: test gate passed (3 targets)`, exit 0.

PR #644, opened 2026-08-17
Direct emission for proven-string and proven-keyword interop targets

yogthos:
Direct emission for interop calls whose target is proven a string or a keyword at compile time, plus three `jolt build` fixes that fell out of a repro along the way.

## What it does

`(.m target arg*)` with a target proven to be a string emits the `jolt-string-method` arm's body inline instead of routing through `record-method-dispatch` with a `jolt-vector` rest-args allocation. The proof comes from two seams: a `^String` hint on the target (symbol tag, let/param binding, or string literal), and the types pass proving `:str` with no hint in the source. The same treatment is applied to `^clojure.lang.Keyword` targets. Chez-only, since the natives live in `java/natives-str.ss` which the Gambit boot excludes; any method or arity outside the fixed tables falls back to the generic dispatch.

Alongside that, `jolt build`'s require-graph scan now walks each closure file's forms for host class references so a lib-provided class (`java.util.Locale`, `ZonedDateTime`) pulls its provider's install namespace into the binary. Three related build bugs came out of that repro: unloaded provider namespaces were re-analyzed against empty refer tables, lazy stdlib was skipped as runtime-owned, and the per-ns record builders emitted out of order. `manifest-check.sh` now gates every boot prologue against the runtime manifest rather than the three that were already pinned.

## Measured

`make test` green on this branch (79 ci targets + 3). Seed re-minted and converged.

The string half delivers what it claims. Interleaved against a `main` release binary, 1M calls each, target hoisted:

| call | main | this branch | JVM |
|---|---|---|---|
| `(.indexOf ^String s 46)` | 140 ns | **23 ns** | 10 ns |
| `(.startsWith ^String s "so")` | 142 ns | **17 ns** | 31 ns |
| `(.substring ^String s 1 4)` | 236 ns | **23 ns** | 17 ns |

That is 6-10x on those call sites, landing at or ahead of the JVM.

## What it does not do, and two things worth fixing before merge

On the honeysql test suite (`jolt -X:test:jolt` in a honeysql checkout, against `clj -X:test`) the end-to-end numbers are:

| | wall clock | vs JVM |
|---|---|---|
| `clj -X:test` | 6.88 s | 1.0x |
| jolt `main` (d77aefaf) | 62.12 s | 9.0x |
| jolt this branch | 61.14 s | 8.9x |

So the suite moves 1.6%, which matches the branch's own note that its cost is the clause walk and not interop. That is fine as far as it goes, but two specific things in here do not hold up:

**The keyword half is dead code for the program it was written for.** The commit motivates it as "honeysql's `kw->sym` is `(.sym ^clojure.lang.Keyword k)` on its `:clj` branch". jolt's reader advertises `:bb` (`reader.ss:619`, `'("jolt" "bb" "clj" "default")`) and honeysql writes `#?(:bb … :clj (.sym ^clojure.lang.Keyword k))` with `:bb` first, at all three `.sym` sites (`sql.cljc:499`, `578`, `598`). jolt takes the `:bb` branch and never reaches the interop. Measured: `kw->sym` is 122 ns on `main` and 123 ns on this branch, against 18 ns on the JVM. The optimization is correct and gated, it simply never fires here.

**The "unresolved oddity" in the build commit has a plain cause.** The message says jolt's lazy `map` realizes the per-ns emit lambdas out of list order. `map` in `build.ss` is Chez's, and R6RS leaves the order of application unspecified — Chez really does apply out of order:

```
$ chez --script -e '(map (lambda (x) (set! log (cons x log)) x) (list 1 2 3 4 5))'
application order: (5 3 4 1 2)
```

The fix is right, but the diagnosis in the comment is not, and the same hazard applies to every other side-effecting `map` in the host rather than just the two sites patched here.

## Review notes

- `infer`'s new `:host-call` arm claims it "only adds, never removes" a `:target-type`, but it unconditionally `assoc`es `:target-type :str` over whatever the analyzer stamped. `toString` and `hashCode` are in both the string and keyword tables, so a `:kw` stamp overwritten by a `:str` inference would emit the wrong body. It is not reachable today without a wrong hint, but the code should match its comment.
- `hint-of` returning `:str`/`:kw` is documented as inert because "every `:hint` consumer equality-checks `:struct`". `passes/inline.clj:77` tests truthiness and propagates the hint. Benign, and arguably desirable, but the comment is wrong.
- Widening `nm` to `clojure.string` in `call-ret-type` also exposes `num-ret-fns` and `vector-ret-fns` to `clojure.string` names. No overlap today; a per-ns lookup would make that safe by construction.
- The comment "chez? is bound before t: `target` (the host predicate) must not be shadowed yet" describes a hazard that no longer exists, since the binding was renamed to `t`.
- `keyword-direct-emit` splices `t` two or three times (`sym`, `toString`, `hashCode`). Sound only because the analyzer stamps `:target-type` exclusively for symbol and literal targets, so `t` is always an identifier or constant. Worth stating, since a future `:hint` on a non-atomic node would silently double-evaluate.
- The last commit is titled `tests` but adds the whole keyword feature. Worth a real message.
- The keyword bodies were checked line by line against `records-dispatch.ss:152-166` and match, including `getNamespace`'s `(or … jolt-nil)` and the `hashCode` `+ #x9e3779b9`.


yogthos:
## Where the honeysql gap actually is

Following up on the 1.6% suite result, I profiled the hot path to find what would close the gap. Everything below is `jolt -M:test:jolt` against `clj -M:test` in the same honeysql checkout, so the only variable is the runtime.

`honey.cache-test` is 34.3 s of jolt's 61.1 s (56% of the run), against 2.4 s of the JVM's 6.9 s:

| section | JVM | jolt | ratio |
|---|---|---|---|
| Uncached, simple, embedded | 1309 ms | 21527 ms | 16.4x |
| Cached, simple, embedded | 241 ms | 2824 ms | 11.7x |
| Uncached, complex, mixed | 822 ms | 9774 ms | 11.9x |
| Cached, complex, mixed | 19 ms | 183 ms | 9.5x |

That section is 100k `honey.sql/format` calls on a 3-key map. Apportioning one call (215 µs on jolt, 19 µs on the JVM):

| | jolt | JVM | ratio |
|---|---|---|---|
| `format` whole | 215 µs | 19.2 µs | 11.2x |
| `format-dsl`, options prebound | 197 µs | 11.1 µs | 17.8x |
| clause walk, `(k m)` only (92 keys) | 6.7 µs | 1.25 µs | 5.4x |
| clause walk, **+ `kw->sym` fallback** | **88.1 µs** | 2.05 µs | **43x** |
| `format-entity :a.b` | 6.5 µs | 0.46 µs | 14.2x |
| `sql-kw :select` | 0.57 µs | 0.09 µs | 6.3x |

`clause-order` is 92 entries, and `format-dsl`'s reduce does, for every clause that is *not* in the statement map, `(get leftover (kw->sym k))` — a freshly allocated symbol used as a map key. Adding that fallback to the walk costs 81 µs on jolt and 0.8 µs on the JVM. It is ~40% of a `format` call on its own.

Drilling in, the cause is not the walk, it is that **symbols are second-class in the equality and hash fast paths**:

| | jolt | JVM |
|---|---|---|
| `(hash :select)` | 23.5 ns | 9.5 ns |
| `(hash 'select)` | **82.5 ns** | 22.5 ns |
| `(= :select :select)` | 11.0 ns | 9.5 ns |
| `(= 'select 'select)` | **173.0 ns** | 12.0 ns |
| `(get m :nope)` | 38.5 ns | 16.5 ns |
| `(get m 'nope)` (4-key map) | 99.0 ns | 24.5 ns |
| `(get m 'nope)` (30-key map) | **309.0 ns** | 12.0 ns |

Both have the same shape in the source. `keyword-t` is `(fields ns name khash)` with a cached hash, and `jolt-hasheq` answers keywords on line 3 of its fast-path `cond` (`hasheq.ss:473`). `symbol-t` is `(fields ns name meta)` with no cached hash, so every symbol hash falls through to the `jolt-hasheq-arms` registry walk and then the legacy `jolt-hash-arms` walk. `jolt=2` is the same story: its fast paths are fixnum, flonum, and `eq?`, so two distinct symbol records walk the whole `jolt-eq-arms` registry before reaching `jolt=2-base`, where symbols are the 5th `cond` arm and compare with `equal?` on the ns plus `string=?` on the name.

Three fixes, in value order, none of them in this PR's scope:

1. **Give `symbol-t` a cached `khash` field and a `jolt-hasheq` fast arm, mirroring `keyword-t`, and hoist the symbol arm in `jolt=2` above the registry walk.** This is the 40%-of-`format` item. Note that `intern-symbol-string` pools the ns/name strings, which tempts an `eq?` comparison — but `values.ss` documents that pool as an unlocked check-then-set that can produce two string objects for one name, so `eq?` would be wrong under that race. Either keep `string=?` and just hoist the arm, or make the pool race-free first.

2. **Invoking a collection in head position.** `(m k)` is 258.5 ns against 38.5 ns for `(get m k)` and 10.5 ns on the JVM — 25x. `(:select m)` with a literal keyword is 32.5 ns, so there is a fast path for a literal keyword head but not for a map head or a keyword-valued local (`(k m)`, 70 ns), which is exactly the shape the clause walk uses.

3. **`format-entity`, 6.5 µs against 0.46 µs.** Not yet broken down; the candidates are the three dynamic-var reads plus destructuring, the `some->`/`re-find` on `quoted-always`, and `sql-kw`.

## On the 30x figure

I could not reproduce it. Measured just now on this machine:

| | wall clock | vs JVM |
|---|---|---|
| `clj -X:test` | 6.88 s | 1.0x |
| jolt v0.7.9 (installed) | 65.46 s | 9.5x |
| jolt `main` (d77aefaf) | 62.12 s | 9.0x |
| jolt `perf/dispatch` | 61.14 s | 8.9x |
| `bin/jolt` (seed path, this branch) | 47.67 s | 6.9x |

So the gap is already under the 10x bar, and has been since at least v0.7.9 — the 30x number looks like it predates the earlier interop-dispatch campaign that is already on `main`. The three items above are what stand between 8.9x and something closer to 4-5x.

One oddity worth noting separately: `bin/jolt` off the seed is **faster** than the release binary on this workload, 47.7 s against 61.1 s. That is a 28% penalty for the AOT release build and I have not chased it. If it is real rather than thermal, it is probably worth more than any of the three items above.

yogthos:
## Correction to the wall-clock numbers above

The end-to-end timings in my comment above were taken while a `make test` run was competing for CPU, which inflated every jolt figure and the JVM baseline unevenly. Re-measured on a quiet machine, best of 2-3 runs each:

| | wall clock | vs JVM |
|---|---|---|
| `clj -X:test` | 5.15 s | 1.0x |
| jolt `main` (d77aefaf) | 44.66 s | 8.6x |
| jolt `perf/dispatch` | 43.59 s | **8.4x** |
| `bin/jolt` seed path, this branch | 46.92 s | 9.1x |

Two things change. The branch's effect on the suite is ~2.4% rather than 1.6%, and the overall ratio is 8.4x rather than 8.9x — both conclusions unchanged, still under the 10x bar and still not what this PR moves.

The third thing changes properly: **the `bin/jolt`-is-faster-than-the-release-binary claim was an artifact.** Quiet, the release binary is 43.6 s and the seed path is 46.9 s, i.e. the AOT build is 7% faster, which is the expected direction. Please disregard that last paragraph; there is no AOT regression to chase.

The per-call breakdown is unaffected — those benches are single-process and CPU-bound, and re-running them quietly reproduced within noise. Restated with the quiet JVM numbers:

| | jolt | JVM | ratio |
|---|---|---|---|
| `format` whole | 209.9 µs | 19.0 µs | 11.0x |
| `format-dsl`, options prebound | 196.5 µs | 11.0 µs | 17.9x |
| clause walk, `(k m)` only (92 keys) | 6.7 µs | 0.90 µs | 7.4x |
| clause walk, **+ `kw->sym` fallback** | **87.6 µs** | 2.95 µs | **29.7x** |
| `kw->sym` x92 | 11.95 µs | 0.75 µs | 15.9x |
| `format-entity :a.b` | 6.43 µs | 0.39 µs | 16.5x |
| `sql-kw :select` | 0.57 µs | 0.09 µs | 6.3x |

| | jolt | JVM | ratio |
|---|---|---|---|
| `(hash :select)` | 24.0 ns | 18.5 ns | 1.3x |
| `(hash 'select)` | **82.5 ns** | 30.0 ns | 2.8x |
| `(= :select :select)` | 11.0 ns | 6.0 ns | 1.8x |
| `(= 'select 'select)` | **177.5 ns** | 35.0 ns | **5.1x** |
| `(get m k)` | 38.5 ns | 10.0 ns | 3.9x |
| `(sm k)` map as fn | **266.0 ns** | 7.0 ns | **38x** |
| `(k sm)` keyword-valued local as fn | 71.0 ns | 6.5 ns | 10.9x |
| `(:select sm)` literal keyword as fn | 33.0 ns | 9.5 ns | 3.5x |

The three recommendations stand, and the ordering is if anything clearer: symbol `=`/`hash` off the fast path, collection-in-head-position dispatch, and `format-entity`.

yogthos:
## Review fixes and the three runtime fixes are in

Four commits on top of the branch. `make test` green on the final state (79 ci targets + 3), seed re-minted and converged.

**Suite result: 43.59 s → 31.61 s against the JVM's 5.18 s. 8.4x → 6.1x.**

| | wall clock | vs JVM |
|---|---|---|
| `clj -X:test` | 5.18 s | 1.0x |
| jolt `main` | 44.66 s | 8.6x |
| branch as reviewed | 43.59 s | 8.4x |
| **branch now** | **31.61 s** | **6.1x** |

### Review fixes (`97da1a31`)

`infer`'s `:host-call` arm no longer overwrites a `:target-type` the analyzer stamped, so "adds, never removes" is what the code does and not just what the comment says. `call-ret-type` gets one `case` arm per namespace, so a bare name can no longer be read against another namespace's table — `clojure.core` and `clojure.string` genuinely disagree on `replace` and `reverse`. The `hint-of` comment is corrected about `passes/inline.clj` testing truthiness. The stale `target`-shadowing comment is gone, `keyword-direct-emit`'s multi-splice invariant is written down, and the note about jolt never reaching honeysql's `.sym` (jolt advertises `:bb`, honeysql orders `:bb` first) is recorded at the emission site.

Two I could not do as asked. Amending the `tests` commit message was blocked, since that commit is already pushed and reachable from this PR — the proper description for it is in `827b9c28`'s message instead, and the commit is still titled `tests`. And I found one more thing while in there rather than just reviewing it: **the string and keyword emit heads were never added to `rt-emitted-names`**, which the INVARIANT directly above that set requires of any new bare Scheme head. A local named `str-trim`, `keyword-t-name` or `render-piece` would have shadowed the emitted call. Fixed in `827b9c28` with unit rows that shadow three of them.

### 1. Symbols get a cached hash and fast paths (`891fbcc9`)

`keyword-t` has carried its hash in the record since it was written; `symbol-t` used a per-thread weak-eq hashtable keyed by the symbol object. That inverted for the case that matters most — a symbol built for one lookup and dropped, which is what `(get m (symbol (name k)))` does — it MISSED and then INSERTED, so the table grew by an entry per call and every one had to be traced and collected. `format-dsl` does that 92 times per `format` call. `symbol-t` has a lazily-filled `khash` field now, and both `jolt-hasheq` and `jolt=2` answer symbols up front instead of behind two registry walks.

The equality compare tries `eq?` on each half before `string=?`, since `intern-symbol-string` pools the ns and name strings. The `string=?` stays: that pool is documented as an unlocked check-then-set that can hand out two objects for one name, and `eq?` alone would make `(= 'foo 'foo)` false for the pair that lost.

| | before | after | JVM |
|---|---|---|---|
| `(hash 'select)` | 82.5 ns | **27.0 ns** | 30.0 |
| `(= 'select 'select)` | 177.5 ns | **44.0 ns** | 35.0 |

Moving symbols onto those fast paths moves them under the arm-registry invariant, and `values-test` caught it — two assertions recorded the old state and are inverted rather than deleted. No arm in the tree claimed symbols, so nothing was silently skipped in the process.

### 2. Lookup-shaped callables in `jolt-invoke1/2` (`5f03411e`)

A keyword, symbol, map, set or vector in head position went through variadic `jolt-invoke`, costing far more than the lookup it decays to: a rest list, `(length args)`, an `apply`, and an arity-name string built whether or not the arity is wrong.

| | before | after | JVM |
|---|---|---|---|
| `(m k)` map head | 256.0 ns | **43.0 ns** | 7.0 |
| `(k m)` keyword local head | 71.0 ns | **41.5 ns** | 6.5 |

A literal `:k` head was already 32.5 ns because the back end lowers it straight to `jolt-get`. The local case is the one `format-dsl` uses 92 times per call.

### 3. `format-entity` → proven-StringBuilder emission (`827b9c28`)

`format-entity` turned out not to be about dynamic vars or destructuring. 45% of it was one call, `join "." (map …) parts`, and inside that `honey.sql.util/join` builds with a `StringBuilder` — 500 ns per `.append` against 20 ns on the JVM, where `sb-append!` itself is three `vector-set!`s. A jhost method call hashes the tag string for the method table, hashes the method name for the handler, converts the rest-args vector back to a list for `apply`, and hands it to a shim taking another rest list.

So this extends the PR's own mechanism to StringBuilder, and the proof source is the interesting part: nobody writes `^StringBuilder` on `(let [sb (StringBuilder.)] …)`, so `init-proves-hint` reads the type off the init instead. A local bound directly to a constructor is proven with nothing in the source, which is the shape every string-building loop in Clojure has — honeysql's `join`, `data.json`'s reader, `clojure.pprint`.

| | before | after | JVM |
|---|---|---|---|
| `join "." (map …) ["a" "b"]` | 2150 ns | **530 ns** | 123 |
| `format-entity :a.b` | 4783 ns | **2710 ns** | 493 |
| `format`, whole | 136.2 µs | **122.1 µs** | 19.0 |

A constructor in *target* position is deliberately not a proof source: `append` splices the target twice (append, then fluent return), so the second splice would build a fresh builder. The new unit row caught exactly that, answering `""` instead of `"lit"`.

Gates: 14 new `sbdirect` unit rows, three cross-platform corpus rows, and `build-app/util/sbjoin` — an unhinted let-bound builder in a `reduce` — with `build-smoke` grepping `flat.ss` for the inline `sb-append!`/`sb-str`, asserting no `record-method-dispatch "append"` survives, and checking the separator, empty and single-element cases at runtime. One corpus row leaves `nil` out of the append chain: `(.append sb nil)` is a real reflection ambiguity on the JVM and raises there, while jolt renders `"null"` — a strict superset the generic path already had, pinned in `unit.edn` instead.

### What is left

`format` is still 122 µs against 19 µs, so there is a 6x there and the suite has more in it. The remaining items I measured but did not chase:

- `kw->sym` × 92 is 11.9 µs against 0.75 µs. It is on honeysql's `:bb` branch, so it is `namespace` + `name` + `symbol` per clause — symbol *construction*, not hashing, and `(symbol "nope")` is 46.5 ns against 47 ns on the JVM, so the cost is the three calls around it.
- The `(get leftover sym)` half of the clause walk is still 38.7 µs against 2.95 µs even after the symbol work.
- `alphanumeric?` is 167 ns against 17 ns (a `transduce` chain), `split-by-separator` 287 ns against 70 ns, `sql-kw` 0.57 µs against 0.09 µs.
- `(hash 7)` is 58 ns against 33 ns, despite `hasheq.ss` claiming a hand-inlined murmur fixnum fast path. Worth a look on its own.

yogthos:
## The hash engine, and two new benchmarks

`40e1d4ae`. `make test` green (ci 80 + test 3 — the extra target is the new `hasheq` gate).

### What the Clojure source settled

Reading `Symbol.java` rules out the shortcut I was hoping for: **Clojure does not intern symbols.** `Symbol.intern` is a bare `new Symbol`, and the `_interned` parameter names in its private constructor are vestigial — the strings aren't interned either. So there is no world where a repeat `(symbol "select")` comes back carrying a hash. Clojure caches `_hasheq` per object exactly as jolt now does, and recomputes for a fresh symbol exactly as jolt does. The gap was pure constant factors in our Murmur3, and the arithmetic is lifted unchanged from `Murmur3.java` and `Util.hashCombine`.

### What was actually slow

`rotl32`, by a factor of three — 6.3 ns against `mul32`'s 2.1 ns — and it runs twice per murmur round, so it dominated everything. Two causes: `(remainder n 32)` was a generic procedure call on what is a literal 13 or 15 at every call site, and the shifts and the `ior` were generic `bitwise-*` rather than fx ops. `urs32` and `hash-combine` had the same problem.

The string murmur loop was also still calling the mixers as procedures. `hasheq.ss` had already hand-inlined them for the int and long paths — noting +220 ns from the layered chain — but left strings on what it called the cold path. It is not cold; it is the symbol path. The three mixers now exist as `-flat` macros as well, so the string loop gets the flat chain without a second hand-expansion, and the procedure forms stay for the genuinely cold bignum arms.

| | before | after | JVM |
|---|---|---|---|
| `rotl32` | 6.3 ns | **2.1 ns** | — |
| `urs32` | 3.3 ns | **2.2 ns** | — |
| `hash-combine` | 8.2 ns | **3.9 ns** | — |
| `murmur3-hash-unencoded-chars "select"` | 67.5 ns | **34.4 ns** | — |
| `murmur3-hash-long-flat 7` | 43.1 ns | **22.2 ns** | — |
| `jolt-hasheq 7` | 48.5 ns | **26.4 ns** | 6.0 |
| `jolt-hasheq` of a FRESH symbol | 115.0 ns | **75.7 ns** | 9.0 |

### One of these went too far, and the corpus caught it

`u32` has to stay generic. Alone among the helpers it is the entry point for the cold arms too — `murmur3-hash-long`'s bignum arm masks a 64-bit value, and `double-hasheq` feeds it a double's raw bits — so an unsafe `fxand` there is applied to a non-fixnum, which does not error, it silently answers wrong. It surfaced as `hash double 1.5` and `hash large long` corpus divergences. It also bought nothing: 2.1 ns either way, because the *result* is always a fixnum and Chez's generic `bitwise-and` is already fast for that case. What the fx forms rely on is that output range, and that is now written down at the point of temptation.

### Regression tests

`test/chez/hasheq-test.ss`, new, wired into `make ci` as `hasheq`. It exists because of that near-miss: a hash engine's values are a wire format, and a wrong one fails nothing locally — maps still work, `=` still answers correctly, and nothing breaks until a hash crosses the JVM boundary or a serialized structure is re-read.

52 checks: literal golden values from JVM Clojure for `hashUnencodedChars`, `Symbol`/`Keyword` hasheq, `hashCombine`, fixnums, and the cold double / wide-long / bignum arms the `u32` slip broke; a flat-vs-layered mixer sweep over every length 0..64 and every char class, with astral chars at each alignment (where jolt's `pending` state machine diverges from Java's fixed 2-char stride, since a Chez string holds codepoints and Java's holds UTF-16 units); the symbol khash cache including the with-meta inheritance path; and hasheq's agreement with `jolt=` for distinct symbol records.

I verified it fails on the `u32` regression before committing it clean — six of the new goldens go red.

Worth noting: my first draft of this file had *guessed* golden values for `hashCombine` and fixnum hasheq, and five of them were wrong while jolt was right. They are all measured from the JVM now, which is the only way this file is worth anything.

### Benchmarks

`bench/keyed_lookup.clj` and `bench/string_build.clj`, both registered in `bench/run.sh` and the README table. These are the honeysql-shaped workloads the existing suite could not see:

- `collections` churns large HAMTs, so a per-key hash amortises across 32-way nodes. At 3-8 entries the hash **is** the cost, and a symbol key built for one lookup and discarded pays a full Murmur3 every time. `keyed-lookup` also puts a collection and a keyword-valued *local* in head position, not the literal `(:k m)` the back end lowers directly.
- `transducers` measures its machinery with an arithmetic reducing function. `string-build`'s rf calls a host method instead, which was the most expensive interop shape jolt had.

Both are sensitive to the fixes behind them. Against the pre-fix binary:

| | pre-fix | now | jvm | ratio |
|---|---|---|---|---|
| `keyed-lookup` | 444 ms | **208 ms** | 50 ms | 4.2x |
| `string-build` | 1322 ms | **205 ms** | 46 ms | 4.5x |

### End to end

The honeysql suite moves 31.4 s to 31.2 s. The engine numbers above are large but the suite's remaining cost is not in hashing, so this lands mostly as headroom plus the two gates. `format` is 116.6 µs against the JVM's 19 µs, down from 122.1 µs.

What is left in `format`, measured but not chased: the `(get leftover sym)` half of the clause walk is 36.9 µs against 2.95 µs, `kw->sym` × 92 is 11.9 µs against 0.75 µs (now dominated by `intern-symbol-string` at 21 ns per call and the record allocation, not the hash), `format-entity` is 4.6 µs against 0.39 µs, and `alphanumeric?` is 167 ns against 17 ns.

PR #645, opened 2026-08-17
Require deps

ingydotnet:
- Add `clojurestar.deps/require-deps`, a portable macro that acquires a
  dependency and imports its namespace at runtime.
- Support Maven libraries, single-file Gists, and single-file GitHub sources
  without invoking a JVM.
- Treat Jolt-provided Clojure runtimes as terminal dependencies, avoiding
  unnecessary Clojure, ClojureScript, and transitive spec downloads.
- Align Maven and Gitlibs cache configuration with tools.deps conventions and
  add Jolt-specific environment overrides.
- Make the source launcher reliably use a compatible Chez and resolve script
  paths relative to the invoking project.

## `require-deps`

Literal libspec vectors use require-style syntax without quoting. Quoted
vectors remain supported for compatibility.

```clojure
(require '[clojurestar.deps :refer [require-deps]])

(require-deps
 ["mvn:dev.weavejester/medley@1.10.0/medley.core" :as medley])
```

Libspecs support `:as` and explicit `:refer [...]` imports. Dependencies are
prepared and loaded from left to right, namespace ownership conflicts are
reported as structured errors, and loading a source file restores the caller's
namespace.

The supported source-file coordinate forms are:

```text
gist:<owner>/<id>/<file>
gist:<owner>/<id>/<file>@<revision>
gist:<owner>/<id>/<revision>/<file>

github:<owner>/<repo>/<ref>/<path.clj|cljc>
github:<owner>/<repo>/blob/<ref>/<path.clj|cljc>
```

Equivalent pinned Gist and GitHub forms normalize to the same identity and
cache entry. Full 40-character GitHub commit SHAs reuse persistent cache;
named refs refresh in a new process. Source files must be self-contained and
begin with an `ns` form.

## Resolution and caches

`org.clojure/clojure` and `org.clojure/clojurescript` are supplied by Jolt and
now terminate transitive expansion. Explicitly declared
`org.clojure/spec.alpha` and `org.clojure/core.specs.alpha` dependencies remain
ordinary dependencies.

Maven repository precedence is:

1. `:mvn/local-repo`
2. `JOLT_MAVEN_REPOSITORY`
3. `GRENADINE_MAVEN_REPOSITORY`
4. `~/.m2/repository`

Relative repository paths resolve against the invoking project directory.

Git, Gist, and GitHub source caches use `:gitlibs/dir`, followed by
`JOLT_GITLIBS_DIR`, `GRENADINE_GITLIBS_DIR`, `GITLIBS`, and Jolt's default.
`:cache-dir` remains a compatibility alias for source-file dependencies.
Downloads use temporary files and atomic moves.

## Source launcher

`bin/jolt` now honors `JOLT_CHEZ`, reuses a threaded Chez 10.x provisioned by
the Make build, and only then searches `PATH`. It also preserves the caller's
directory in `JOLT_PWD`, allowing relative script arguments and relative Maven
repository overrides to keep project-relative behavior after the launcher
changes into the Jolt checkout.

## Testing

- `make grenadine grenadinecheck`
- Direct source-mode Maven dependency loading with the rebased launcher
- Individual `:as` and inline `:local/root` dependency-loading checks
- `make unit`: 1316/1318 passed; the same two shared-process failures reproduce
  on an unmodified `origin/main` worktree, while both cases pass individually
- `make smoke`: the release binary builds successfully; the remaining smoke
  run is blocked in the development sandbox by denied access to
  `~/.clojure/deps.edn` and `~/.jolt/aot-cache`


ingydotnet:
requires https://github.com/clojurestar/grenadine/pull/6 mere

yogthos:
Reviewed and verified locally against the branch with the 0.1.7 submodule checked out. CI is green and I am merging this.

The headline feature works end to end. `require-deps` resolved `mvn:dev.weavejester/medley@1.10.0/medley.core` and a `gist:` coordinate on a first run, the gist landed at `~/.jolt/gitlibs/gist/<owner>/<id>/latest/source.clj`, and the relative-script fix works from an arbitrary cwd. The remote-source path looks right to me: fetches go through `jolt.mvn-http/fetch` over cert-verified TLS, land in a `.part` file and atomic-move into place, and the boolean result is checked rather than assumed. Coordinate parsing validates each path segment against a charset, excludes `.` and `..`, requires filenames to match `\.cljc?`, and requires full 40-character SHAs for pins, so there is no way out of the cache root. The Chez probe is also fine: Chez prints `--version` to stdout, the `(threaded?)` probe correctly rejects non-threaded builds, and an unmatched glob falls through harmlessly.

On Clojure becoming terminal, I want to flag what it costs so it is on the record. A project whose only declared dependency is `org.clojure/clojure` can no longer require `clojure.spec.alpha`. On `main` `test/chez/deps-alias/specproj` prints `spec: true false`; on this branch the same project (with its original deps.edn) dies with `Could not locate clojure/spec/alpha.jolt`. That matters because the JVM artifact does depend on spec.alpha and core.specs.alpha, which is why #528 added the substitution in the first place, and libraries do require spec without naming it. We are taking this as a deliberate design decision rather than a bug, so I am merging, but note that no gate covers it any more: the specproj deps.edn and the `grenadine_test` assertion were both rewritten to match, and the conformance manifest hand-lists `@spec.alpha/src/main/clojure` as a source root so it never exercises deps.edn resolution. I will record it in `known-divergences.edn` and the differences docs in a follow-up.

Four things I would still like fixed, none of them blocking:

The four renamed environment variables now fail silently. `JOLT_LOCAL_REPO=/tmp/x` resolves to `~/.m2/repository`, and the changelog advertised that knob as still supported. An ignored env var gives no signal, so a user who deliberately moved their repository will quietly re-download into the default location. Keeping the old names as deprecated fallbacks, or warning when one is set, would cost very little.

`:provided-libs` is dead code. Grenadine 0.1.7 threads it through `expander` and `basis`, but nothing in `jolt-core` passes it, so the Clojure pruning still happens entirely through jolt's own `intrinsic-dep?` in `filter-deps`. Worth either wiring up or saying explicitly that it is groundwork for a later change.

`clojurestar.deps` is no longer dialect-neutral, though the docstring still says it is. The implementation require is reader-conditioned for `:jolt`, `:bb` and `:clj`, but `[grenadine.require-deps :as required]` is unconditional and `require-deps*` calls `implementation/prepare-required!`, which only `jolt.deps` defines. On babashka the namespace will not load at all. Either reader-condition the new pieces or drop the portability claim from the docstring.

`file-arg` breaks Windows absolute paths. `(str/starts-with? x "/")` does not match `C:\script.clj`, which now becomes `./C:\script.clj`. Windows is a shipped platform, so this is a real if narrow break of something that worked before.

Some smaller things:

- The README now carries the "Source code archives omit submodules" note twice, at lines 69-71 and again at 100-102.
- `grenadine_test.clj:837` reports "relative Grenadine repository did not use the project directory", but the assertion it guards exercises the Jolt override in the second positional argument.
- `{:keys [coordinate] alias-symbol :as refer-symbols :refer}` in `require-deps*` is correct, since `:as` and `:refer` are the values rather than the keys, but it reads like Clojure's whole-map `:as`. Three plain bindings would be easier to trust at a glance.
- `load-required!` shadows `clojure.core/identity` with a local, and its check-then-`swap!` is not atomic, so two threads could both load the same coordinate.
- `defonce required-state` sits about 1400 lines above its only users.

PR #651, opened 2026-08-18
Make time report like the reference macro

yogthos:
Noticed while benchmarking honeysql: every `time` line jolt printed differed from Clojure's, and comparing the two outputs side by side was the whole point of the exercise.

    jolt   Elapsed time: 10231 msecs
    clj    "Elapsed time: 1169.953292 msecs"

Two differences, and the first is not cosmetic.

**The clock.** `time` measured with `current-time-ms`, the epoch wall clock in whole milliseconds. So anything faster than a millisecond printed `0 msecs` — useless for exactly the microbenchmarking the macro gets reached for — and a clock adjustment mid-expression could print a negative elapsed. `~/src/clojure/src/clj/clojure/core.clj` measures `(System/nanoTime)`, which is monotonic and fine-grained.

jolt already has that clock: `jolt-mono-nanos`, sitting behind `System/nanoTime`, with a comment explaining it is Chez's `time-monotonic` precisely because a wall-clock-derived one "could run backwards" and timed sub-millisecond intervals as 0. `time` just wasn't using it. This exposes it as `clojure.core/current-time-ns` and measures with it, divided to fractional milliseconds. Keeping it a core fn rather than having the macro reach for `System/nanoTime` keeps the core overlay independent of the java class-statics layer.

The comment above the macro claimed `current-time-ms` was "the host's monotonic clock". It is not — it is `now-millis`. That is gone.

**The output.** The reference prints through `prn`, not `println`, so the line is a quoted string. jolt printed it bare.

After:

    jolt   "Elapsed time: 0.002 msecs"
    clj    "Elapsed time: 0.004583 msecs"

Sub-millisecond intervals are now visible rather than reading `0 msecs`. The remaining difference is resolution, which is host-inherent.

## Tests

Three corpus rows pin the shape, since the figure itself cannot be asserted: the line starts with a quote then `Elapsed time: `, ends with ` msecs` then a quote, the elapsed figure reads back as a number, and `time` still returns its expression's value. All three verified against a real Clojure, which answers identically to jolt.

`make corpus` green (4461/4470, 0 new divergences). The local `make test` run hit one failure — `POLLER-REGISTRATION LOST 1 of 8 readiness registrations in round 29` in the cli smoke gate — which is a race in the io-poller test, unrelated to this change; `make smoke` passes 146/146 on re-run, twice. CI runs the full gate here.


yogthos:
Rebased onto main via a merge (`6e214bf8`) to clear the conflict.

The conflict was **only** in `host/chez/seed/{image,prelude}.ss`, which are generated: this branch reminted them, and so did #650 when it landed. Hand-stitching two generated outputs would be wrong, so I took main's seed and re-ran `make remint` over the merged sources, regenerating it from them. Converged in 2 passes.

The real sources merged with no conflict — the changes are disjoint (`values.ss` / `collections.ss` / `backend_scheme.clj` from main, `30-macros.clj` / `host-static-methods.ss` here).

Verified after the merge that both sides are live in one binary: `time` prints the quoted line, `(= :a :b)` takes #650's keyword fast path, and `(class (seq [1 2]))` is `PersistentVector$ChunkedSeq` from #648. Full gate green — 81 ci targets, 3 test targets.

PR #657, opened 2026-08-18
Measure the read and print scaling gates accurately

yogthos:
Follow-up to #656, which is where these gates first got in the way: `printscaling` failed twice during that work and `readscaling` once. One of the print failures was on a tree with no local changes at all, ratio 8.71 against a ceiling of 8.0, which is what made it worth looking at rather than re-running.

Both gates judge a ratio — quadruple the input, and the cost must not quadruple per element — and both were deciding that from measurements too imprecise to support it.

## The timer

`currentTimeMillis`, while the 1x arm of the print gate renders in 2.4ms. That is a systematic bias, not just noise: on an idle machine the vector arm reported 4.50 or 5.00 across five runs, in discrete steps, where the true figure under `nanoTime` is 4.08. A fifth of the headroom to the ceiling was being eaten by the clock.

## The read gate's window

Two more problems there. It took a single sample per arm, so nothing could absorb one bad measurement. And its sizes were wrong in a way that hid the real story: per-form cost is flat at ~5.0us up to 8000 forms, steps to ~8.2us at 16000, then is flat again (9.46us at 32000, 9.70 at 64000, i.e. linear). The 8000 -> 32000 window straddled that step, so the gate was measuring the step and not the reader — the honest interference-free ratio was ~7.4 against a ceiling of 8.0, and the 2.88 it used to report was an imprecise 1x arm flattering the result. Moving both arms onto the flat side, 16000 -> 64000, measures 4.19-5.32 against a linear expectation of 4.0.

## The change

`nanoTime` in both, best-of several runs per arm instead of one, and re-measure up to three times before failing. Reported ratios are tight now instead of quantized: the print arms sit at 4.07-4.15 and 4.39-4.42 across runs.

The print gate cannot be resized the way the read gate was. Its per-element cost grows across the whole range (record 1.77us at 4000 rising to 5.93 at 64000; vector 0.61 to 1.84), so 16000 -> 64000 measures 9.05 on the vector arm and would fail outright. Its sizes are the best window available and the only remaining lever is taking enough samples to find the floor, so it takes 15.

## Detection power is unchanged

That is the reason to retry rather than just raise the ceiling. Against a deliberately quadratic record render, the ratios are 15.87, 16.14, 16.06 — over the ceiling on every attempt, still failing with exit 1. The read gate also still fails fast on an unambiguous regression: a ratio at or above 12 skips the re-measuring, because at 64000 forms the old drain-and-push-back bug copies ~8e10 chars and retrying it three times would turn a clear failure into a CI timeout.

## What this does not claim

Under controlled CPU contention — 4 and then 12 spinners against 10 cores — the old gates passed six runs out of six. So this is not a fix for ordinary load. It is a fix for imprecise measurement, for judging the wrong window, and for deciding from a single sample. The failures that prompted it were rare; the retry is what turns a rare event into a non-event.

Unloaded, each gate goes from ~0.4s to ~3.7s. That is the cost of sampling enough to be right.

`make test` passes on an idle machine (ci gate 81 targets, test gate 3 targets), with read at 3.87 and print at 4.05 and 4.45, no retries needed.

Two things a reviewer might reasonably ask about, both left alone deliberately. The other scaling gates share the millisecond timer (`hotpath_scaling_test.clj` reports arms of 3ms and 13ms, for instance), so they have the same latent imprecision, but they are passing and I did not want to touch gates I had no failure to point at. And the print gate's per-element cost genuinely rising with n — record 1.77us to 5.93us over this range — may be worth its own look; it is why the ratio sits at 4.4 rather than 4.0.

### DIFF
--- a/test/print_scaling_test.clj
+++ b/test/print_scaling_test.clj
@@ -32,10 +32,13 @@
   (let [r (reduce (fn [r i] (assoc r (keyword (str "k" i)) i)) (->ExtRec) (range n))]
     (count (pr-str r))))
 
+;; nanoTime, not currentTimeMillis. The 1x arm here renders in 2-3ms, so
+;; millisecond resolution quantized it to ±33% — and that error lands straight in
+;; the ratio, since t1 is the denominator.
 (defn- timed [f]
-  (let [t (System/currentTimeMillis)
+  (let [t (System/nanoTime)
         v (f)]
-    [(- (System/currentTimeMillis) t) v]))
+    [(/ (- (System/nanoTime) t) 1e6) v]))
 
 (defn- best-of [k f]
   (reduce min (map first (repeatedly k #(timed f)))))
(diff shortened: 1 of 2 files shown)


yogthos:
CI caught a real problem with the first cut of this, and it is worth recording because it is the same class of mistake the PR is about.

The `complexity` gate failed, not either of the gates I changed. The cause is that CI runs the suite as `make -j$(nproc) test`, so the gates run in parallel — and my first version made these two ~9x more expensive (read 0.41s to 3.7s, print likewise) by reading 16000/64000 forms and taking 15 samples per arm. Holding a core that much longer starved the neighbouring timing gates, and `complexity` has a ceiling of 2.0 over 3ms measurements, so it was the one that tipped: it reported 2.80 for `transient write-few`. My own gates passed while breaking someone else's. I had also justified those 15 samples with load figures that turned out to be measured on a machine polluted by orphaned spinners from my own testing, so that number was never sound.

The fix keeps the accuracy and drops the cost. The read gate now measures 2000 -> 8000 rather than 16000 -> 64000: the point of resizing was to get both arms onto one side of the per-form cost step between 8000 and 16000, and the side *below* it works just as well while being an order of magnitude cheaper. It also measures tighter, 4.15-4.36 against 4.19-5.32 for the region above the step. Print keeps its sizes, which cannot be changed for reasons in the file, and samples 5 instead of 15.

Cost is now what it was before the change: read 0.41s against 0.41s, print 0.44s against 0.34s. Detection power is unchanged — a deliberately quadratic record render still measures 15.69, 16.25, 16.06 and fails on every attempt.

Verified locally with `make -j$(sysctl -n hw.ncpu) test`, mirroring how CI invokes it: read passes at 3.83, print at 4.35 and 4.51, and `complexity transient write-few` comes in at 1.09 against its 2.0 ceiling rather than the 2.80 CI saw.

PR #679, opened 2026-08-20
Honor a ^long/^double return hint on the arglist, not just the defn name

yogthos:
Reopens #678, which GitHub auto-closed when its base branch was deleted on merging #677. Same change, retargeted at `main`.

`(defn f ^long [x] x)` and `(fn ^long [x] x)` declare that arity's primitive return type, and the reference honors both spellings. jolt read the tag only off a `defn`'s **name**, so the arglist spelling was silently dropped:

```clojure
((fn ^long [x] x) 18446744073709551616N)   ; reference 0   jolt 18446744073709551616N
```

`strip-arglist-meta` throws the hint away so fn parsing sees the params — which is right, that is what parsing needs — but its comment went further and asserted the hint "is ignored at runtime", which is the bug rather than a design note. The kind is now read off the **original** form before the strip.

Both reader shapes are covered: the vector carries its metadata directly (what jolt's reader produces — verified), and a reader that lowers `^meta` on a collection to `(with-meta <vec> <meta>)` puts it in the third element. A non-numeric tag (`^String [x]`) still yields nil, so class hints on an arglist stay ignored exactly as before. An explicit `:ret-nhint` already on the arity wins, so a name tag and an arglist tag cannot fight.

## Before / after

| spelling | reference | before | after |
|---|---|---|---|
| `(defn ^long f [x] x)` | 0 | 0 | 0 |
| `(defn f ^long [x] x)` | 0 | **2^64** | 0 |
| `(def g (fn ^long [x] x))` | 0 | **2^64** | 0 |
| multi-arity `(fn (^long [x] x) (^long [x y] …))` | 0 | **2^64** | 0 |
| `(fn ^double [x] x)` ← 5 | 5.0 | **5** | 5.0 |
| `(fn ^String [x] x)` | passthrough | passthrough | passthrough |

Five corpus rows pin these against JVM Clojure, including that a `^String` arglist tag stays inert.

## Known gap, filed separately

An anonymous fn whose call the **inliner** can see through still loses the coercion. Unchanged by this PR, and a different mechanism:

```clojure
((fn ^long [x] x) v)                      ; still uncoerced — IIFE
(let [g (fn ^long [x] x)] (g v))          ; still uncoerced — let-bound
((identity (fn ^long [x] x)) v)           ; still uncoerced
(let [g (fn ^long [x] x)] (fn [] (g v)))  ; correct — inliner cannot see through the nested fn
```

That last row is the tell: the same fn is correct when the call hides behind another fn. So the inliner beta-reduces and splices the body without carrying the arity's return coercion. Fixing it means teaching `inline.clj` to wrap an inlined body in the return coercion, and it needs the *return* variant (`jolt->fx-ret` from #677), not the `:coerce` node's parameter one.

## Gates

`make test` green (ci 81 targets, test 3), `make certify` 0 new / 0 stale, `make selfhost` fixpoint holds, corpus 0 new divergences.


yogthos:
Correction: the "Known gap, filed separately" section in this PR's description is **wrong about the cause**, and I want that on the record here rather than only in the follow-up.

I attributed the lost return coercion to the inliner beta-reducing without carrying it. It is not the inliner. `inline.clj` does carry it (`rbody (if ret (coerce-node ret rbody0) rbody0)`), and none of the cases I listed even had a var in call position for that path to fire on.

The actual cause, fixed in #683: `#()` rebuilds its whole body to substitute the `%`-params, and the rebuild dropped reader metadata from every nested collection literal. A `^long` return hint lives on the arglist **vector**, so the hint went with it.

What misled me is worth naming: I used `#(…)` as the probe harness in every one of those examples, so "IIFE", "let-bound" and "via `identity`" looked like three independent observations converging on an inliner theory when they shared one confound. Varying the *nesting shape* instead — top level, inside a `defn`, inside a `def`'d fn, immediately applied, and passed to a fn are all correct; only `#()` is wrong — is what actually located it.

It was also much broader than type hints. `#(meta ^{:a 1} [1 2])` read `nil` where the reference gives `{:a 1}`, for vectors, maps and sets at any depth. So this PR's table is still accurate for what it claims, but the gap it points at was a different and larger bug than described.

PR #692, opened 2026-08-20
Add exact-width scalar FFI types

casselc:
## Summary

- add exact-width FFI scalar types `:int8`/`:i8`, `:int16`/`:short`, `:uint16`/`:ushort`, `:int32`, and `:uint32`
- keep the runtime memory type table and compiler signature type table in lockstep
- document native byte order and C default promotions after `:varargs`
- add C-backed coverage for memory access, native arguments and returns, callbacks, aliases, boundaries, and fail-closed unknown types

## Motivation

Jolt currently cannot faithfully describe fixed-width C integers narrower than `:int` or distinguish fixed 32-bit integers from platform C integer aliases. Using `:int` for fields such as a C `short` gives the wrong size and prevents signatures and native-memory layouts from sharing one exact type vocabulary.

The existing `:uint8`/`:u8`/`:byte` aliases and all existing FFI behavior remain unchanged.

## Validation

After rebasing onto current `main`:

- `make -j1 ffi`: 17/17 and 51/51
- `make -j1 selfhost`: checked-in seed matches the rebuild
- `make -j1 unit`: 1387/1387
- `git diff --check`
- shell syntax check and strict C helper compilation with `-Wall -Wextra -Werror`

Before the final current-`main` rebase, the complete serial `make -j1 test` gate passed all 81 CI targets and all 3 top-level test targets on the v0.7.18 base. The intervening upstream commits are confined to the InputStream surface; the focused FFI, self-host, and expanded unit gates were rerun after rebasing.


yogthos:
It looks good to me, let me know when you're ready for review, and can merge it in.

yogthos:
and will have this in the next release shortly, thanks for the pr :)

Issue #383, opened 2026-07-17
add nix flake


yogthos:

jasalt:
Investigated Nix flake support but it got too complicated to me as I'm rather new to Nix (learning to use it with devenv.sh recently).

Draft PR (https://github.com/jolt-lang/jolt/pull/694) has basic Linux support and LLM review suggesting slight change on Jolt `mvn_http.clj` to improve multiplatform support which I can't judge how feasible it is.

PR #694, opened 2026-08-21
Add initial Nix flake support for Linux x86_64

jasalt:
Provides reproducible devShells, standalone packages.default, nix run app, formatter, and recursive submodule support.

Validated on x86_64 Linux:

- nix flake check --no-build
- nix develop --command bin/jolt -e '(+ 1 2)'
- nix build and built binary execution
- nix run . -- -e '(+ 40 2)' → 42

Related https://github.com/jolt-lang/jolt/issues/383

### DIFF
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -9,9 +9,19 @@ permissions:
 
 jobs:
   test:
-    name: Linux x86_64
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Linux x86_64
+            os: ubuntu-latest
+            system: x86_64-linux
+          - name: macOS arm64
+            os: macos-latest
+            system: aarch64-darwin
     steps:
       - uses: actions/checkout@v5
         with:
@@ -23,7 +33,7 @@ jobs:
       - name: Check the flake and formatting
         run: |
           nix flake check --no-build --no-write-lock-file
-          nix build .#formatter.x86_64-linux --no-write-lock-file
+          nix build .#formatter.${{ matrix.system }} --no-write-lock-file
           ./result/bin/nixfmt --check flake.nix
 
       - name: Build and smoke-test the packaged Jolt
@@ -35,6 +45,18 @@ jobs:
           trap 'rm -rf "$tmp" result' EXIT
           env -i HOME="$tmp/home" XDG_CACHE_HOME="$tmp/cache" PATH=/usr/bin:/bin \
             "$out/bin/jolt" -e '(+ 40 2)' | grep -Fx '42'
+
+      # jolt.mvn-http's built-in macOS candidates are the Homebrew OpenSSL
+      # paths, and the runner image ships Homebrew OpenSSL — which would let
+      # the fetch below pass without the Nix-provided OpenSSL ever loading.
+      # Remove the Homebrew entries so success can only come through the
+      # wrapper's JOLT_OPENSSL_LIBDIR.
+      - name: Drop Homebrew OpenSSL from the built-in candidate paths (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          sudo rm -f /opt/homebrew/opt/openssl@3 \
+            /opt/homebrew/lib/libcrypto.dylib /opt/homebrew/lib/libssl.dylib
+          sudo rm -rf /usr/local/opt/openssl@3
 
       - name: Verify packaged dependency resolution
         run: |
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -62,12 +62,16 @@ jobs:
           export REAL_GIT="$(command -v git)"
           cat > "$shim/git" <<'GIT'
           #!/usr/bin/env bash
-          case "${1:-}" in
-            clone|fetch)
-              echo "unexpected Make bootstrap: git $1" >&2
-              exit 99
-              ;;
-          esac
+          # Git accepts options before its subcommand (for example,
+          # `git -C DIR fetch`), so inspect every argument rather than $1.
+          for arg; do
+            case "$arg" in
+              clone|fetch)
+                echo "unexpected Make bootstrap: git $*" >&2
+                exit 99
+                ;;
+            esac
+          done
           exec "$REAL_GIT" "$@"
           GIT
           chmod +x "$shim/git"
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -36,7 +36,7 @@ jobs:
           env -i HOME="$tmp/home" XDG_CACHE_HOME="$tmp/cache" PATH=/usr/bin:/bin \
             "$out/bin/jolt" -e '(+ 40 2)' | grep -Fx '42'
 
-      - name: Verify the packaged HTTPS trust store
+      - name: Verify packaged dependency resolution
         run: |
           set -euo pipefail
           out="$(nix path-info .#jolt --no-write-lock-file)"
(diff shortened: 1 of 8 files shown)


jasalt:
Review (gpt-5.6-sol) suggests that this gets tricky with MacOS support.
___

- [P1] flake.nix:29-35 — standalone package omits runtime dependencies.     
  The package closure contains neither OpenSSL, git, nor unzip, although    
  Jolt requires them for deps.edn resolution. Consequently, nix run depends 
  on undeclared host tools and Maven/Git dependencies fail on a minimal     
  NixOS system. Wrap jolt with the required PATH and OpenSSL library path,  
  or otherwise include these runtime dependencies hermetically.             
                                                                            
- [P1] flake.nix:89-92 — OpenSSL in the development shell is not dynamically
  discoverable.                                                             
  pkgs.openssl adds linker flags but no runtime                             
  LD_LIBRARY_PATH/DYLD_LIBRARY_PATH. Jolt loads OpenSSL through FFI; Linux  
  uses bare sonames, while macOS only tries Homebrew paths. Thus Maven      
  dependency fetching fails in a pure Nix environment—especially on macOS,  
  where no Nix store path is attempted.                                     
                                                                            
- [P2] flake.nix:21,39-43 — package and embedded Jolt versions disagree.    
  The derivation is named jolt-66bf430, but the built executable reports    
  jolt dev. The source copied into the Nix store has no Git metadata, so    
  build-jolt.ss cannot run git describe. Export JOLT_VERSION from the       
  derivation version during the build.

norbot7865:
Re: https://github.com/jolt-lang/jolt/pull/694#issuecomment-5366849172 — all three findings check out against the source. I dug into how `jolt.mvn-http` actually loads OpenSSL; some detail that sharpens the picture, plus a proposed resolution.

## The two P1s are one root cause

`jolt.mvn-http` `dlopen`s OpenSSL lazily at runtime (`stdlib/jolt/mvn_http.clj`), so Nix never adds it to the binary's RPATH (nothing is *linked*), and the built-in candidate lists can't see Nix store paths:

- **Linux:** bare sonames (`libcrypto.so.3`, …) resolve only via ldconfig/`LD_LIBRARY_PATH` → fail on minimal NixOS and in a pure dev shell on any host lacking `/usr/lib/libcrypto.so.3`.
- **macOS:** only explicit Homebrew paths are tried. The bare `libcrypto.dylib` name is *deliberately* excluded (`mvn_http.clj:22-24`) because `/usr/lib/libcrypto.dylib` is a protected Apple image that kills the process with an uncatchable SIGABRT on `dlopen`. Consequence: **no `DYLD_LIBRARY_PATH`-style wrapper can fix macOS** — a code change is required. `LD_LIBRARY_PATH` would half-fix Linux but leaks into user-loaded natives, so it's not the clean answer either.

### Proposed fix: `JOLT_OPENSSL_LIBDIR` override + `wrapProgram`

Teach `mvn_http.clj` to prepend `$JOLT_OPENSSL_LIBDIR/<per-OS soname>` to the candidate lists — the caller supplies a directory, Jolt keeps the filename mapping:

```clojure
(defn- lib-candidates [libdir names fallbacks]
  (into (if (and libdir (not (str/blank? libdir)))
            (mapv #(str libdir "/" %) names)
            [])
        fallbacks))
```

Flake side:

```nix
nativeBuildInputs = [ pkgs.chez pkgs.pkg-config pkgs.xxd pkgs.makeWrapper ];
buildInputs = [ pkgs.lz4 pkgs.zlib pkgs.ncurses pkgs.openssl ]
  ++ pkgs.lib.optionals pkgs.stdenv.hostPlatform.isLinux [ pkgs.libuuid ]
  ++ pkgs.lib.optionals pkgs.stdenv.hostPlatform.isDarwin [ pkgs.libiconv ];

postFixup = ''
  wrapProgram "$out/bin/jolt" \
    --prefix PATH : ${pkgs.lib.makeBinPath [ pkgs.git pkgs.unzip ]} \
    --set-default JOLT_OPENSSL_LIBDIR ${pkgs.openssl.out}/lib
'';
```

And `JOLT_OPENSSL_LIBDIR = "${pkgs.openssl.out}/lib";` in the dev shell — on Linux too, for the NixOS case above.

Why this shape over `substituteInPlace`-patching the candidate lists in `postPatch`:

1. **One code path covers all consumption modes.** The release binary bakes `mvn_http.clj` as an in-heap source string (`build-jolt.ss` bakes all of `jolt-core/`+`stdlib/`; `build.ss:578` AOT-emits the CLI's require closure), and the env var is read at namespace load in the user process — so the baked binary, `bin/jolt` dev mode, and the dev shell all honor it identically. Source patching reaches the baked binary but leaves every dev checkout broken, and is fragile string surgery on a moving list.
2. **House style.** `JOLT_CHEZ`, `JOLT_PWD`, `JOLT_NO_USER_DEPS`, `JOLT_AOT_CACHE`, `JOLT_VERSION`, `JOLT_BIN` already exist; `System/getenv` is live and unfiltered by default. One caveat worth documenting: app builds that set `JOLT_BAKE_ENV_ALLOWLIST` must allowlist `JOLT_OPENSSL_LIBDIR` (`host/chez/java/host-static-methods.ss:916`).
3. `--set-default` (not `--set`) preserves user override; safe here because a Nix store path is a real OpenSSL, unlike the bare-name hazard on macOS.
4. The env var is generic, not Nix-specific (MacPorts `/opt/local`, Guix, nonstandard Homebrew prefixes all benefit) — happy to split the `mvn_http.clj` change into its own PR if you'd rather land it ahead of the flake.
5. It's testable without network/OpenSSL: candidate construction as a pure function fits `test/mvn_http_test.clj` (the existing `mvnhttp` gate is pure-function tests); `make httpsfetch` remains the opt-in end-to-end check.

Optionally also expose `packages.jolt-unwrapped` (binary + link deps only) with the wrapped `packages.jolt` as `default` — keeps a minimal package available while making `nix run` correct.

## P2 (version mismatch)

Confirmed: `build-jolt.ss:70-74` prefers `$JOLT_VERSION`, then `git describe`, then `"dev"` — and the flake's `src` has no `.git`, so it falls through. One line fixes both the derivation name and `--version` output:

```nix
env.JOLT_VERSION = self.shortRev or "dev";
```

## macOS validation (recommended)

macOS is where every divergence lives (protected system libcrypto, Homebrew-only candidates), so Linux+Nix alone can't prove the fix. Suggested steps on an Apple Silicon Mac without Homebrew OpenSSL (or with `/opt/homebrew/opt/openssl@3` temporarily renamed):

1. **Pre-fix:** `nix build`, then
   `env -i HOME="$HOME" ./result/bin/jolt -e '(require (quote jolt.mvn-http)) (jolt.mvn-http/fetch "https://repo1.maven.org/maven2/org/clojure/math.combinatorics/0.2.0/math.combinatorics-0.2.0.pom" "/tmp/j.pom")'`
   → expect graceful failure (`fetch` fails soft by design). If it *succeeds*, it silently used Homebrew — exactly the ambient-state dependence to eliminate.
2. **Post-fix:** same command → `:ok` + non-empty pom; `nix path-info -r result | grep openssl` shows OpenSSL in the closure; `otool -L result/bin/jolt` shows no Homebrew linkage.
3. **End-to-end:** temp project with one `:mvn/version` dep and one `:git/sha` dep → `env -i PATH=/usr/bin:/bin result/bin/jolt run -m app.core` exercises OpenSSL + git/unzip wrapping together.
4. **Override:** `JOLT_OPENSSL_LIBDIR=/bogus` → graceful failure, proving `--set-default` semantics.
5. **Dev shell:** `nix develop` → `make mvnhttp` and `make httpsfetch` pass without Homebrew, and `bin/jolt` dev mode fetches too (the case source-patching can't fix).

Also worth expanding `systems` to `aarch64-linux` / `x86_64-darwin` — nixpkgs' Chez supports all four, and the platform conditionals (`libuuid`/`libiconv`) are already in place.

---

_Follow-up review performed with the kimi-k3 LLM model, with all claims verified against the source tree at 51630218 (this PR's head) — file/line references above are from that checkout, not model recall._

jasalt:
Changing this PR to Linux X86_64 only focus for now.
MacOS support could land as separate PR.

jasalt:
Left documentation out for now so can be decided how it would be best done if this goes forward. Nix flake usage could be documented in relevant README sections (run, build, ...) or as it's own section, something in-line of:

> Nix flake is supported for Linux x86_64. `nix run . -- -e '(+ 1 2)'` runs the packaged Jolt, `nix build .#jolt` produces `result/bin/jolt`, and `nix develop` provides the source-build toolchain. The development shell uses the flake-locked makeplus/makes source, so `make` does not bootstrap it from GitHub.

> - `nix run . -- -e EXPR`: run the packaged Jolt on Linux x86_64.
> - `nix build .#jolt`: build the Linux x86_64 package (`result/bin/jolt`).
> - `nix develop`: enter the source-build shell; its makeplus/makes input is locked
>   by the flake, so `make` does not bootstrap it from GitHub.

Mac specific changes would follow naturally after it gets support.

jasalt:
Green flake test run https://github.com/jasalt/jolt/actions/runs/32472237732

yogthos:
Thanks — this was in good shape, so I took it the rest of the way rather than splitting a follow-up. Pushed three commits on top:

- `25dd3660` fixes a real bug the macOS work surfaced: since the scoped-native loader landed, `jolt.ffi/load-library` swallowed dlopen failures and returned nil regardless, so the candidate fallback in `jolt.mvn-http` never advanced past a missing first entry (and http-client's try/catch probe couldn't see failures either). It raises again now, with regression rows in the ffi gate.
- `49ec6ed0` adds the `JOLT_OPENSSL_LIBDIR` override discussed above (named per the review comment). Read at fetch time in `ensure-native!`, so baked binaries and restored images honor the runtime env; entries are absolute paths into the named dir, so the macOS protected-stub hazard doesn't apply. Verified live on an arm64 Mac: lsof shows the fetch served from the override dir, and bogus dirs fall back gracefully.
- `8fcba79f` turns the flake claim into x86_64-linux + aarch64-darwin: libuuid/libiconv become per-platform (matching build.ss link flags), the wrapper sets `JOLT_OPENSSL_LIBDIR` instead of `LD_LIBRARY_PATH`, and CI is a two-job matrix. The macOS job deletes the runner's Homebrew OpenSSL before the dependency-resolution check so it can only pass through the Nix OpenSSL. Intel macOS stays out for now — releases cross-build it from the arm64 runner and there's no equivalent Nix path to validate.

Full local gate is green (`make test`, 81 targets, plus the opt-in `make httpsfetch` under four OpenSSL-dir scenarios). Will merge once both CI jobs pass.

PR #698, opened 2026-08-22
Add declarative FFI struct layouts

casselc:
## Summary

- add literal, data-only `[:struct ...]` descriptors through `jolt.ffi/layout`
- derive native size, alignment, and nested field offsets from lexically scoped Chez `define-ftype` declarations
- add `layout-size`, `layout-alignment`, `field-offset`, `read-field`, and `write-field`
- reject empty or malformed descriptors and unsupported field kinds at compile time
- compare flat, padded, and nested layouts against C ABI witnesses

Descriptors are immutable data rather than entries in a mutable registry. This initial surface supports fixed-size scalar fields and nested structs; arrays, unions, bitfields, packing, and recursive descriptors remain unsupported.

## Stack

This is 1 of 3 in the FFI stack:

1. This PR: declarative struct layouts
2. #699: by-value aggregate calls
3. #700: scoped allocation helpers

GitHub native stacks do not support cross-fork branches, so the three PRs target `main` and are linked explicitly.

## Validation

- rebased onto official v0.7.22 (`8990efe2`)
- pinned Chez 10.4.1 wrapper preflight
- focused FFI: bindings 19/19, exact widths 51/51, layouts 25/25; public API test passed
- layout public API passed from source and a rebuilt standalone binary
- Chez seed reminted to a two-pass byte fixpoint
- Gambit seed regenerated and `gambitseedcheck` passed
- serial `make -j1 test`: 82 CI targets and 3 test targets passed
- unit gate: 1394/1394
- `git diff --check`
- [fork CI](https://github.com/casselc/jolt/pull/51) passed on Linux x86-64 and macOS arm64


yogthos:
Nice, this should make FFI more ergonomic. :)

PR #710, opened 2026-08-23
Preserve Windows absolute paths through project resolution

casselc:
## Problem

Running Jolt from source on native Windows with a drive-absolute `JOLT_PWD`
fails before user code starts. For example, a project at
`D:\a\app` is read as:

```text
D:\a\app/D:\a\app/deps.edn
```

The same leading-slash assumption also affects absolute `:local/root` and
cache paths during dependency expansion. A drive-rooted or UNC path can be
prefixed with the project directory a second time.

This surfaced independently in casselc/jolt-tcp#6 against released Jolt
v0.8.1. Its native Windows x86_64 and ARM64 jobs both fail in
`jolt.deps/read-deps-file` before any TCP test executes.

## Root cause

There are two path boundaries, and each recognized only a leading `/` as
absolute:

- the Chez `java.io.File` host layer resolves paths used by `slurp`,
  `getAbsolutePath`, and `isAbsolute`; and
- the Clojure dependency resolver makes roots absolute while walking the
  project graph.

Fixing only the dependency classifier does not repair a drive-absolute
`JOLT_PWD`, because the host layer must first open the project's `deps.edn`.

## Change

- Recognize Windows drive-rooted paths with either separator and UNC/device
  paths without touching the filesystem.
- Recognize drive-rooted and UNC paths as absolute for filesystem access,
  `getAbsolutePath`, and `File.isAbsolute`.
- Keep a single-leading-separator path non-absolute, as on the JVM, while
  resolving it against `JOLT_PWD`'s drive rather than the launcher's drive.
- Preserve dependency root spelling for drive and UNC/device inputs, and
  resolve current-drive-rooted inputs against the declaring base's drive.
- Reject ambiguous drive-relative forms such as `C:project` with structured
  `{:path ... :kind :drive-relative}` data rather than manufacturing a path
  under the launch directory.
- Add a native Windows source-mode gate that passes a real drive-absolute
  `JOLT_PWD` for a temporary project through `-Spath` and asserts that the
  unique root declared by its `deps.edn` resolves. Stock code misses that file
  after doubling its path and falls back to `src`, so this marker makes the
  check non-vacuous. The same native step checks that `\rooted.txt` is
  non-absolute but resolves from `user.dir`'s drive.

The native source gate uses PowerShell so MSYS path conversion cannot turn the
drive-rooted input into a leading-slash form and make the reproducer vacuous.

## Validation

All local Jolt commands used Chez Scheme 10.4.1.

- dependency expansion: 65/65 passed;
- host unit suite: 1657/1657 passed;
- workflow lint: passed;
- `git diff --check`: passed.

The added Windows assertions and source-mode invocation fail on the exact
current-upstream base `05b573a98e4b834e65e924d69d84df53a94b771f`:
drive/UNC `File.isAbsolute` is classified incorrectly, and a drive-absolute
`JOLT_PWD` is duplicated while opening `deps.edn`.

Hosted run 33947866847 job 101257111802 passed on candidate
`f7f5dd31251e0b04dc4f8f377a7779f89b932379`: 65/65 dependency-expansion
cases, the temporary project's unique `marker-src` root through native
source-mode `-Spath`, and the current-drive-rooted `File` assertion all passed.
The same run's aggregate job 101257111669 passed. Flake run 33947866809 also
passed on Linux x86_64 and macOS arm64.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -159,10 +159,16 @@ jobs:
       # Use PowerShell so JOLT_PWD retains its native drive-rooted spelling.
       # The MSYS shell can translate it before the runtime sees the value,
       # which would make this regression check vacuous.
-      - name: Resolve from a native absolute project directory
+      - name: Resolve a fixture from a native absolute project directory
         shell: powershell
         run: |
-          $env:JOLT_PWD = (Resolve-Path '.').Path
+          $project = Join-Path $env:RUNNER_TEMP 'absolute-path-project'
+          $marker = Join-Path $project 'marker-src'
+          New-Item -ItemType Directory -Force -Path $marker | Out-Null
+          Set-Content -NoNewline -Encoding ascii `
+            -Path (Join-Path $project 'deps.edn') `
+            -Value '{:paths ["marker-src"]}'
+          $env:JOLT_PWD = (Resolve-Path $project).Path
           if ($env:JOLT_PWD -notmatch '^[A-Za-z]:[\\/]') {
             throw "expected a drive-absolute JOLT_PWD, got: $env:JOLT_PWD"
           }
@@ -170,8 +176,8 @@ jobs:
           $resolved = (& $scheme --script host\chez\cli.ss -Spath 2>&1 | Out-String)
           Write-Host $resolved
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
-          if (-not $resolved.Contains('jolt-core')) {
-            throw "resolved path did not contain the project's jolt-core root"
+          if (-not $resolved.Contains('marker-src')) {
+            throw "resolved path did not contain the fixture's declared marker root"
           }
           $probe = '(let [u (System/getProperty "user.dir") p "\\rooted.txt" f (java.io.File. p)] (and (false? (.isAbsolute f)) (= (str (subs u 0 2) p) (.getAbsolutePath f))))'
           $rooted = (& $scheme --script host\chez\cli.ss -e $probe 2>&1 | Out-String).Trim()
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -180,7 +180,9 @@ jobs:
             throw "resolved path did not contain the fixture's declared marker root"
           }
           $env:JOLT_EXPECTED_ROOTED = $env:JOLT_PWD.Substring(0, 2) + '\rooted.txt'
-          $probe = '(let [p "\\rooted.txt" f (java.io.File. p)] (and (false? (.isAbsolute f)) (= (System/getenv "JOLT_EXPECTED_ROOTED") (.getAbsolutePath f))))'
+          # Avoid string literals in the native `-e` argument: Windows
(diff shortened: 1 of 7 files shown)


yogthos:
Reviewed. The classification work is right — drive-rooted with either separator, UNC/device, `C:x` and `\x` correctly *not* absolute, and rejecting drive-relative rather than manufacturing `D:/base/C:path` is the right call. Behaviour on POSIX is unchanged (`native-path-kind-for` with `windows? false` reduces to exactly the old `starts-with "/"`), which I checked rather than assumed.

I pushed one follow-up (a35bbbf) for a performance regression the change introduces, plus two small things in `io.ss`.

## `(sa-os-family)` is not cached, and this puts it on every filesystem touch

`jfile-path-absolute?` asks it per call, and `project-relative` — via `jfile-fs` — sits under every `.exists`, `.isDirectory`, `slurp`, `listFiles` child and `getAbsolutePath`. `sa-os-family` allocates a fresh string through `symbol->string` on every call and then walks it with a `substring` per position, five needles deep:

```
.isAbsolute  x200k   141ms  ->  355ms
.exists      x50k    116ms  ->  167ms
```

The note on `sa-probed-os-family` already argues for caching — *"consulted from hot paths and the answer cannot change while the process runs"* — but the cache is on the **probe**, which only a `pb` tag reaches, so `ta6le`/`ta6nt`/`tarm64osx` re-derived every time. Memoizing at `sa-os-family` puts both numbers back under baseline (136ms / 113ms) and helps every other caller too. `hostprops` still passes (68 checks) since it pins the table through `sa-os-family-for-tag`, which I didn't touch.

## `project-relative` could make a rooted path relative

With `JOLT_PWD` unset, `jolt-user-dir` is `"."`, and the `else` arm produced `".\x"` from `"\x"` — a rooted path turned into a relative one, which is worse than leaving the OS to resolve it against the process drive. `jolt.deps/root-relative-for` already keeps that arm and `(absolute true "." "\project")` pins it, so the Scheme and Clojure classifiers disagreed on the one input neither can resolve. Aligned, including the UNC-base arm `io.ss` was missing.

Also: `jfile-abs` classified the path and then handed it to `project-relative`, which classified it again — `project-relative` already answers the absolute case with itself, so only the empty-path case needed keeping.

## Left alone on purpose

`jolt.deps/windows?` re-probes `os.name` on every `abspath`. I measured it at 405ns; against tens to hundreds of roots per resolution, next to the file I/O that resolution is made of, that's noise — and a memo there is a real hazard, since a build-time `abspath` call could bake the build host's answer into an AOT image. Not worth it.

## One thing worth knowing about

There are now two implementations of the same classifier — `jfile-path-absolute?`/`windows-root-relative?` in Scheme and `native-path-kind-for` in Clojure — and no gate cross-checks them, since the Windows arms are unreachable on the Linux/macOS runners. They had already drifted on the `"."` base before this commit. Unavoidable across the host boundary as far as I can see, but a future change to either rule has to touch both, and nothing will say so.

Gates run locally: `depsunit` 65/65, `unit` exit 0 (`io-file-base` 3/3), `hostprops` 68/68, `adaptercheck`, `statlayout` 11/11.

PR #722, opened 2026-08-24
java: support Thread.getId

casselc:
## Summary

Add `Thread.getId` compatibility using Jolt's stable numeric identity for the
running thread. This is the surface SCI reaches during Context initialization;
it does not claim JVM Thread semantics beyond Jolt's thread model.

## Tests

- Direct unit coverage for numeric, repeatable current-thread identity.
- Direct coverage that simultaneously live threads observe distinct IDs.
- Both cases fail on pristine upstream and pass with this change.
- `make unit`
- `make threadsafety`
- `make fibers`
- `make selfhost` (exact fixpoint)

All local Jolt commands used Chez 10.4.1.

### DIFF
--- a/host/chez/java/io.ss
+++ b/host/chez/java/io.ss
@@ -1268,6 +1268,7 @@
 (register-host-methods! "thread"
   (list (cons "getContextClassLoader" (lambda (self) the-classloader))
         (cons "getName" (lambda (self) "main"))
+        (cons "getId" (lambda (self) (get-thread-id)))
         ;; no reified call stack (jolt does TCO, so caller frames are erased) — an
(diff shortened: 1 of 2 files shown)


yogthos:
Merged past the red `test` check: the failure is the `complexity` gate's `transient write-few` section — a timing-ratio check that its own comments document as noise-prone on shared runners. The diff (one `getId` registration) can't affect transient freeze scaling. Verified locally on a branch combining #721+#722+#723+#724: full `make ci` green (83 targets), including `complexity: passed`, unit 1414/1414, and the new `interruptnest` gate.

Issue #290, opened 2026-07-02
AOT build: binary fails at startup with 'variable var-deref is not bound' when a namespace is required at runtime (works under joltc run)


tompassarelli:
## Environment

- jolt commit `51dec5fd2c7e99be16b0f58f8e395391f8223ac5`
- Chez Scheme 10.4.1
- Linux x86_64 (NixOS)

## Summary

A namespace that is only required at runtime (the `requiring-resolve` pattern — dynamic `require` + `resolve`, not part of the static `ns` graph) works fine under `joltc run`, but the AOT-built binary fails at startup with:

```
Unhandled exception: Exception: variable var-deref is not bound
```

Both release and `--dev` builds fail identically.

## Minimal repro

`deps.edn`:

```clojure
{:paths ["src"]}
```

`src/repro/extra.clj`:

```clojure
(ns repro.extra)

(defn greet [s]
  (str "Hello, " s "!"))
```

`src/repro/main.clj`:

```clojure
(ns repro.main)

(defn -main [& args]
  (require 'repro.extra)
  (println "required ok"))
```

Commands (from the project root):

```
$ joltc run -m repro.main
required ok

$ joltc build -m repro.main -o repro-bin
jolt build: wrote ./repro-bin
$ ./repro-bin
Unhandled exception: Exception: variable var-deref is not bound

$ joltc build -m repro.main -o repro-dev --dev
$ ./repro-dev
Unhandled exception: Exception: variable var-deref is not bound
```

## Expected

The built binary behaves like `joltc run`: the runtime `require` loads `repro.extra` and the program prints `required ok`.

## Actual

The binary dies at startup with `variable var-deref is not bound`.

## Notes

- The generated `repro-bin.build/flat.ss` *does* contain `(define (var-deref ns name) (var-cell-root (jolt-var ns name)))`, so the primitive exists at the compiled program's top level — it looks like the runtime compiler invoked by the dynamic `require` evals its emitted code in an environment that can't see the program's top-level definitions.
- Since `repro.extra` isn't statically reachable, it isn't compiled into the image; run from a different cwd the binary instead fails with `Could not locate repro/extra.clj (or .cljc) on the source roots`, i.e. the binary falls back to loading/compiling the source from disk at runtime, and that runtime-compile path is what hits the unbound `var-deref`.
[...]

PR #756, opened 2026-08-27
fix(build): app images must not inherit the CLI's AOT closure as preloaded

burinc:
## The problem

A namespace that jolt's own CLI closure defines, and that is neither in the
runtime image nor in `host/chez/stdlib-fasl-manifest.txt`, ends up with every
var it defines **interned but unbound** in a binary written by `jolt build`.

Two namespaces are in that position today:

| namespace | affected vars |
|---|---|
| `jolt.ffi` | `layout-size`, `layout-alignment`, `field-offset`, `read-field`, `write-field`, `errno`, `errno-message` |
| `jolt.mvn-http` | `fetch`, `fetch*` |

`jolt.ffi`'s host-provided primitives (`alloc`, `free`, `read`, `write`,
`sizeof`, `null`, `ptr->string`) are fine — they come from
`host/chez/java/ffi.ss`. It is exactly the `stdlib/jolt/ffi.clj` half that goes
missing.

`jolt run` masks it completely, because there the source is on disk and gets
compiled at require time. Only a built binary fails, and it fails at the
**call**, not the build.

```clojure
(ns repro (:require [jolt.ffi :as ffi]))
(def L (ffi/layout [:struct [[:x :float] [:y :float]]]))
(defn -main [& _]
  (println (:size L))            ; 8    — fine, map key
  (println (ffi/sizeof :float))  ; 4    — fine, host primitive
  (println (ffi/layout-size L))) ; unbound in a built binary
```

```
$ jolt run -m repro          →  8 / 4 / 8
$ jolt build -m repro -o r && ./r
8
4
Unhandled exception: Attempting to call unbound fn: #'jolt.ffi/layout-size
```

## Why

`bld-emit-cli-aot` bakes `jolt.main`, `jolt.deps` and their on-demand require
closure into the CLI boot image, emitting `(ldr-mark-loaded! …)` for each. Its
own comment names that closure:

> `jolt.main + jolt.deps and their on-demand Clojure require closure
> (clojure.string, clojure.edn, jolt.mvn-http, jolt.ffi, grenadine.*)`

At app-build time `bld-boot-loaded` snapshots `loaded-ns`, and
`bld-require-closure` skips anything in it as already-in-the-image. That is
right for the **runtime** image, which every app shares. It is wrong for the
**CLI's own** AOT closure, because an app image is a different image and
carries none of it.

`clojure.string` and `clojure.edn` sit in that same closure and survive only
because they are independently in the runtime image. That is the whole
discriminator, and it is why the fasl manifest is not the lever — adding
`jolt.ffi` there changes nothing for app builds (I tried it first).

## The change

Record *why* a namespace is preloaded, not merely that it is.

- `loader.ss` gains `ldr-cli-aot-ns` plus `ldr-mark-cli-aot!` / `ldr-cli-aot?`.
- `bld-emit-cli-aot` emits `(ldr-mark-cli-aot! …)` alongside its existing
  `(ldr-mark-loaded! …)`.
- `bld-require-closure` no longer treats a CLI-AOT-only namespace as preloaded,
  so it is emitted into the app image like any other lazy stdlib.

20 lines across two files. `make-devboot.ss` shares `bld-emit-cli-aot`, so the
dev boot cache gets the same treatment.

## Verification

macOS 15 / AArch64, Chez from Homebrew, baseline built from a pristine detached
`aafa0fc3` (confirmed 0 diff, 0 fix markers) so the before/after is real:

```
before  jolt.ffi/layout-size      unbound in a built binary
        jolt.mvn-http/fetch       unbound in a built binary
after   both bound and callable; layout-size / field-offset / read-field /
        write-field all round-trip in AOT
```

Neighbours unchanged before and after — `clojure.test`, `jolt.fs`,
`jolt.process`, `jolt.time.base`, `clojure.string`, `clojure.edn` stay bound.

```
make ffi             green — 45/45 layout, 18/18 aggregate, 39/39 scoped
stdlib-fasl smoke    11 passed, 0 failed
app binary           +128 KB (the cost of actually carrying the two namespaces)
```

Also exercised end to end by a real consumer: a 600-function generated raylib
binding built with `jolt build` and run, with 35/35 struct sizes and 167/167
field offsets agreeing with clang.

## One thing that is not mine

`make certify` fails on this branch — and identically on an unmodified
`aafa0fc3` build, with one NEW divergence:

```
[host-interop / reducible dispatch] 2-arity reduce over an eduction stays on the seq path
  (reduce + (eduction (map inc) [1 2 3]))
  jolt-expected="9"  JVM raised ClassCastException
```

Pre-existing and unrelated, but it means `make test` is currently red on main.
Happy to split that into its own issue.

## Related

#290 — same works-under-`run` / fails-in-AOT shape, different cause.

### DIFF
--- a/host/chez/build.ss
+++ b/host/chez/build.ss
@@ -686,7 +686,10 @@
                   ;; defines above already installed every var. Without this the
                   ;; loader sees an unmarked ns and recompiles it from source on the
                   ;; first command that enters jolt.main/-main (run/build/version).
-                  (put-string out (string-append "(ldr-mark-loaded! " (ei-str-lit name) ")\n")))))
+                  (put-string out (string-append "(ldr-mark-loaded! " (ei-str-lit name) ")\n"))
+                  ;; ...and record that this ns is preloaded only because the CLI
+                  ;; image defines it, so an app build does not inherit the claim.
+                  (put-string out (string-append "(ldr-mark-cli-aot! " (ei-str-lit name) ")\n")))))
             ordered))
         (lambda ()
           (set-optimize! #f)
@@ -990,7 +993,9 @@
             (let ((file (find-ns-file name)))
               (when (and file
                          (or (not (ldr-install-file? file))
-                             (not (hashtable-ref bld-boot-loaded name #f))))
+                             (not (hashtable-ref bld-boot-loaded name #f))
+                             ;; preloaded only in the CLI image, not in an app's
+                             (ldr-cli-aot? name)))
                 (dfs (append (bld-ns-class-providers file) (bld-ns-requires file)))
                 (set! order (cons (cons name file) order)))))
           (dfs (cdr ns)))))
--- a/host/chez/loader.ss
+++ b/host/chez/loader.ss
@@ -1059,6 +1059,19 @@
     (else (parameterize ((aot-dep-sink #f) (io-file-read-sink #f)) (load-jolt-file file)))))
 
 ;; Mark a namespace as loaded in both the host hashtable and the *loaded-libs* ref.
+;; Namespaces defined by the CLI's OWN AOT closure (bld-emit-cli-aot bakes
+;; jolt.main, jolt.deps and their on-demand requires into the CLI boot image and
+;; marks each loaded). They really are preloaded in the jolt process — but an app
+;; image written by `jolt build` is a DIFFERENT image and carries none of them,
+;; so the app build must not skip them as "already in the image". Without this,
+;; a ns that is in the CLI closure and neither in the runtime image nor the
+;; stdlib-fasl manifest — jolt.ffi, jolt.mvn-http — has every var it defines
+;; interned but UNBOUND in a built binary, while `jolt run` masks it by
+;; compiling the source at require time.
+(define ldr-cli-aot-ns (make-hashtable string-hash string=?))
+(define (ldr-mark-cli-aot! name) (hashtable-set! ldr-cli-aot-ns name #t))
+(define (ldr-cli-aot? name) (hashtable-ref ldr-cli-aot-ns name #f))
+
 (define (ldr-mark-loaded! name)
   (jolt-with-mutex ldr-tbl-mu (hashtable-set! loaded-ns name #t))
   (ldr-libs-update! (lambda (s) (pset-conj s (jolt-symbol #f name)))))


burinc:
Filed #757 with the reproduction and the scope analysis (which namespaces break and why the fasl manifest is not the lever). This PR is the fix for it.

burinc:
The `make certify` failure mentioned above is now #758 — unrelated to this PR, and it reproduces on a pristine `aafa0fc3`.

burinc:
CI note: the `test` job here failed on the **complexity** gate, not on anything this PR touches —

```
FAIL complexity transient write-few: persistent! is rebuilding the whole map
     ratio 2.12 (ceiling 2.0)
```

That is a wall-clock ratio gate on transient `persistent!`. This PR only changes which namespaces the app-build emits, plus one hashtable in the loader; neither runs during `make complexity`.

Locally on this branch the same gate passes comfortably:

```
complexity transient write-few   18ms at n, 20ms at 4n, ratio 1.11 (ceiling 2.0)
complexity: passed
```

2.12 vs 1.11 looks like a loaded shared runner. The `tests` workflow has also gone red on `main` itself recently (`3277b37d`), so the gate seems timing-sensitive in CI generally.

Worth noting what *did* pass in that same CI run, since it is the part relevant here: `build smoke: passed (release + optimized + direct-link + tree-shake + compiler+core shake + data-reader + no-main + optional-native + …)`, plus both `Linux x86_64` and `macOS arm64` builds. I cannot re-run the job myself — a maintainer re-run would confirm.

(The certify failure I mentioned earlier is #758 and is separate again.)

yogthos:
Reviewed and verified locally on macOS 15 / AArch64.

Reproduced the bug against a build of current main: `jolt.ffi/layout-size` unbound in a built binary, fine under `jolt run`. Both named namespaces are fixed by this — `jolt.mvn-http/fetch` and `fetch*` too.

I went looking for the failure mode this could have: the third OR-clause is `(ldr-cli-aot? name)` on its own, so I expected `clojure.string` and `clojure.edn` — which you correctly identify as sitting in the same closure — to start being emitted redundantly into every app image. They do not, and the reason is worth writing down: `bld-emit-cli-aot` only bakes what is not already in the runtime image, so it never marks them. The marks are exactly the 11 namespaces the CLI had to carry itself:

```
grenadine.expander  grenadine.pom  grenadine.require-deps  grenadine.version
grenadine.xml  jolt.deps  jolt.deps.edn  jolt.deps.ext  jolt.ffi  jolt.main
jolt.mvn-http
```

Confirmed on an app requiring `clojure.string` + `clojure.edn`: the app image emits only the app namespace, before and after. So the clause is precisely scoped, and recording *why* a namespace is preloaded is the right shape — the boot snapshot conflated two different images.

One gap: there is no regression test. This is the "works under `run`, fails only in AOT" class that has bitten this repo before (#290, and the cljc-conditional truncation before that), and it is invisible to every existing gate. I have a `test/chez/ffi-app` fixture that calls the Clojure half of `jolt.ffi` plain and tree-shaken, alongside the `fs-app`/`process-app` ones — it fails on an unfixed build and passes on yours. I tried to push it here but cannot write to your fork from this environment, so it lands as a follow-up PR immediately after this merges, along with the changelog entry.

Also confirming your closing note: `make certify` failing on the eduction row is not yours, and #759 handles it. Thanks — good diagnosis and a clean fix.

PR #759, opened 2026-08-27
certify: allowlist the 2-arity reduce-over-eduction row as :flaky

### LINKED ISSUE #758: certify: the new 2-arity reduce-over-eduction corpus row fails locally (green on CI) — needs its sibling's :flaky entry
**Corrected after filing:** I first wrote that `make test` is red on `main`.
That is wrong — CI is **green** at `aafa0fc3`. The failure is
**environment-dependent**: it fails on a local `make test` and passes on CI,
which is the split the allowlist comment below predicts exactly. That makes it a
worse trap than a plainly red gate, not a lesser one, because a contributor
whose JVM picks `IReduceInit` cannot get a green local gate and CI will not tell
them why.

`make test` fails **locally** on `main` at `aafa0fc3` (verified via
`git ls-remote`, 2026-08-27). The `certify` gate fails, and because it fails the
targets after it never run — so nothing after `certify` in the gate is evidence
of anything.

```
=== NEW divergences (not in allowlist) — gate FAILS ===
  [host-interop / reducible dispatch] 2-arity reduce over an eduction stays on the seq path
      actual: (reduce + (eduction (map inc) [1 2 3]))
      jolt-expected="9" JVM raised ClassCastException: class clojure.core.Eduction
      cannot be cast to class clojure.lang.IReduce
make[2]: *** [certify] Error 1
```

## This is the case the allowlist already documents

`test/conformance/known-divergences.edn` carries this, for a *sibling* row:

> `(reduce + (eduction …))` with no init. jolt folds it; the JVM's answer is
> UNSPECIFIED, so this row is `:flaky` rather than a divergence either way.
> `clojure.core.protocols/CollReduce` is extended to both `IReduceInit` and
> `Iterable`, and an Eduction implements both, so `find-protocol-impl` reduces
> … identity hash codes: every local run here picks `IReduceInit` and CI picks
> `Iterable`, so allowlisting it as a plain divergence fails one of the two.

```clojure
{:suite "seq / lazy over infinite", :label "eduction reduces",
 :category :permissive, :flaky true}
```

The row that is now failing is the same expression and the same
unspecified-dispatch problem, but it was added to the corpus without the
`:flaky` treatment its sibling has:

```clojure
;; test/chez/corpus.edn:1072
{:suite "host-interop / reducible dispatch"
 :label "2-arity reduce over an eduction stays on the seq path"
 :expected "9"
 :actual "(reduce + (eduction (map inc) [1 2 3]))"
 :portability :jvm}
```

Added in `36105184` ("reduce/into/vec dispatch to IReduce(Init); reduce-kv
consults IKVReduce").

`:expected "9"` encodes the `Iterable`/seq outcome, which is only one of the two
the JVM may pick — and the label ("stays on the seq path") asserts the path
rather than merely the value.

## Which branch a JVM takes, measured

macOS 15 / AArch64, JDK 25.0.2, Clojure 1.12.3, isolated project
(`{:deps {org.clojure/clojure {:mvn/version "1.12.3"}}}`, nothing else):

```
8 fresh JVMs, same expression:   ClassCastException  ×8
(reduce + 0 (eduction …))        => 9   (3-arity is well-defined: IReduceInit)
(instance? IReduce      e)       => false
(instance? IReduceInit  e)       => true
```

So on this host the `IReduceInit` branch wins consistently and the row fails
consistently. It is stable *per environment*, not per run.

What flips it is another `CollReduce` extender on the classpath. In a project
whose `deps.edn` pulls `cnuernber/dtype-next`, the identical expression returns
`9`. That is worth knowing because it makes a casual check look fine — I chased
that for a while before spotting it.

For reference, jolt answers `9`.

## Suggested fix

Give the new row the same treatment as its sibling — allowlist it as
`:flaky true` (or `:permissive`), since the JVM side is genuinely unspecified.
Alternatively drop `:expected "9"` in favour of the 3-arity form
`(reduce + 0 (eduction (map inc) [1 2 3]))`, which is well-defined via
`IReduceInit` and still exercises the dispatch the row is there to test.

Happy to send a PR for whichever you prefer.

## Note

Found while validating #756 / #757 — I needed a clean baseline and hit this on
an unmodified tree, so it is unrelated to that work. Confirmed by reproducing it
on a pristine detached `aafa0fc3` with no local patches.

burinc:
Fixes #758.

## The row

`test/chez/corpus.edn:1072`, added in `36105184`:

```clojure
{:suite "host-interop / reducible dispatch"
 :label "2-arity reduce over an eduction stays on the seq path"
 :expected "9"
 :actual "(reduce + (eduction (map inc) [1 2 3]))"
 :portability :jvm}
```

## Why it cannot be pinned

The allowlist already explains this, one row above, for the sibling case:

> `clojure.core.protocols/CollReduce` is extended to both `IReduceInit` and
> `Iterable`, and an Eduction implements both, so `find-protocol-impl` reduces
> `pref` over `(supers Eduction)` — a hash SET — and `pref` keeps whichever of
> two unrelated interfaces it sees first. Iterable wins and the row answers 9;
> IReduceInit wins and the 2-arity casts to `IReduce`, which an Eduction is
> not, and it raises ClassCastException.

`seq / lazy over infinite / eduction reduces` carries `:flaky true` for exactly
that. The new row is the same expression reached from the other side — it
asserts the Iterable outcome — and did not get the same treatment.

## The change

Six lines: one allowlist entry plus a comment saying which sibling it matches
and why.

## Measured

From a pristine `aafa0fc3`, macOS 15 / AArch64, JDK 25.0.2, Clojure 1.12.3:

| | exit | entries | known | NEW | stale |
|---|---|---|---|---|---|
| without | **1** | 106 (1 flaky) | 106/107 | **1** | 0 |
| with | **0** | 107 (2 flaky) | 107/107 | **0** | 0 |

`0 stale` both ways, so the entry masks nothing that was otherwise reported.

## It is an environment split, not a plain failure

This project's CI is **green** at `aafa0fc3` — I checked before writing this,
having got that exact claim wrong on the issue first time round and had to
correct it. It fails on a local `make certify` (8/8 runs here). Which is the
more awkward shape for a contributor: no clean local gate, and CI offering no
explanation.

For what it is worth, an ambient `deps.edn` changes the answer too — in a
project pulling `cnuernber/dtype-next`, a second `CollReduce` extender, the
same expression returns `9`. Consistent with the hash-order explanation, and a
good reason not to trust a casual one-liner check here.

## One thing worth a second opinion

The label asserts the **seq path** specifically, and `:expected "9"` encodes
only the Iterable outcome. If pinning that path is what the row is *for*, the
more honest fix may be the 3-arity form:

```clojure
(reduce + 0 (eduction (map inc) [1 2 3]))   ;; => 9, well defined via IReduceInit
```

which still exercises the dispatch the suite is testing, without a label that
is true only half the time. I do not know the intent behind `36105184` — happy
to switch this PR to that instead.

### DIFF
--- a/test/conformance/known-divergences.edn
+++ b/test/conformance/known-divergences.edn
@@ -776,6 +776,12 @@
    ;; Iterable, so allowlisting it as a plain divergence fails one of the two.
    {:suite "seq / lazy over infinite", :label "eduction reduces", :category :permissive, :flaky true}
 
+   ;; Same expression, same unspecified dispatch, reached from the other side:
+   ;; this row asserts the Iterable outcome ("stays on the seq path", :expected
+   ;; "9"), so it fails wherever find-protocol-impl happens to pick IReduceInit.
+   ;; Added in 36105184 without the :flaky its sibling above already carries.
+   {:suite "host-interop / reducible dispatch", :label "2-arity reduce over an eduction stays on the seq path", :category :permissive, :flaky true}
+
    ;; with-open closes anything carrying a :close fn; the JVM demands a member
    ;; named close and raises on a map. line-seq and file-seq are the same shape:
    ;; jolt duck-types the reader and takes a path string, where the JVM insists on


yogthos:
Reviewed. Merging as-is, and answering the open question: keep `:flaky`, do not switch to the 3-arity form.

I confirmed the mechanism independently rather than take the sibling comment's word for it. `find-protocol-impl` folds `pref` over `(disj (supers c) Object)` filtered to extenders — a `PersistentHashSet`, so iteration order is identity-hash order — and `pref` on two unrelated interfaces just keeps the first:

```
(filter #(get (:impls CollReduce) %) (disj (supers (class (eduction (map inc) [1 2 3]))) Object))
;; => [java.lang.Iterable clojure.lang.IReduceInit]
```

Stable 8/8 here (Iterable first, answers 9), reproducibly the other way for you. Environment-stable, spec-unspecified — exactly the sibling's situation, so exactly the sibling's treatment.

On the 3-arity alternative: it would drop the coverage. The row exists for what `36105184` made interface-scoped — 2-arity reduce drives a declared `IReduce` only, and an Eduction declares just `IReduceInit`, so it must fall to the seq path. `(reduce + 0 …)` goes through `IReduceInit` deliberately; it exercises a different arm and would stop catching a regression where jolt starts treating `IReduceInit` as `IReduce` in the 2-arity case (which raises, so the row does catch it today).

The label is also better than it looks, because the two gates are not the same gate. `host/chez/run-corpus.ss` keeps its own label allowlist and never reads `known-divergences.edn`, so `:flaky` relaxes only the JVM oracle in `certify`. `make corpus` still pins jolt at 9. The label describes jolt's behaviour, which is deterministic; only the oracle's is not, and that is the half being excused.

Verified in a full gate with this and four other PRs merged together: `108 entries (2 flaky); 106 of 106 divergences known, 0 NEW, 0 stale` — inert here, as you measured, and masking nothing.

PR #787, opened 2026-08-30
Fix source-mode builds omitting CLI-preloaded namespace code

jasalt:
## Summary

Source-mode `jolt build` could omit the Clojure definitions of a dependency from a generated executable or shared-library image when that namespace had already been loaded by the CLI process but was not part of the runtime image inherited by the generated artifact.

For `jolt.ffi`, compiled layout values were present, but Clojure vars such as `jolt.ffi/layout-size` were interned with unbound roots. Calling one of those vars from the generated image failed with an unbound-function error.

The issue was discovered while validating Raylib aggregate FFI in the companion Android experiment:

- [RAY-012 experiment](https://github.com/jasalt/jolt-android-experiment/tree/master/experiments/RAY-012-android-aggregate-abi)
- [Reproduction and validation evidence](https://github.com/jasalt/jolt-android-experiment/blob/master/experiments/RAY-012-android-aggregate-abi/README.md)

## Root cause

Before calling `build-binary`, the source-mode CLI loads `jolt.main` and its dependency closure. That process loads lazy Jolt namespaces such as `jolt.ffi`.

`build-binary` previously initialized `bld-boot-loaded` by copying the process's current `loaded-ns` table. The build therefore treated CLI-loaded namespaces as though they were already present in the runtime image inherited by the new executable or library.

Because `jolt.ffi` was already loaded in the build process, the namespace-load hook did not record it. The static require-closure walk also skipped it because the incorrect snapshot classified it as preloaded. Its Clojure definitions were consequently not emitted into the generated image.

This affected the source-mode build path. `jolt run` masks the problem by loading the namespace implementation from source.

## Changes

- Snapshot the namespaces actually present in the runtime image before CLI namespace loading.
- Use that immutable baseline when deciding which Jolt-owned namespaces must be emitted into an application or library image.
- Add a source-mode executable regression using an application that requires the CLI-preloaded `jolt.ffi`.
- Extend the shared-library smoke fixture with an exported `point_size` function that calls `jolt.ffi/layout-size`; the C driver resolves the export through `dlopen` and `jolt_lookup`.

## Validation

- The source-mode executable regression produces:
  - `FFI-APP 8 4 4 2.5 true`
- The source-mode shared-library smoke test passes:
  - `add(2,3) = 5`
  - `point_size() = 8` through `dlopen` and `jolt_lookup`
- With the patch applied to the pinned Jolt revision, RAY-012 passes on an API-35 x86_64 emulator running the ARM64 library through Berberis translation:
  - `jolt.ffi/layout-size` bound: `1`
  - `Color` layout size: `4`
  - complete layout and aggregate matrix: `1`

No unsafe pointer workaround was introduced.

_____
LLM generated content (gpt-5.6-sol).
Beyond my skillset to verify validity end-to-end.


yogthos:
Reviewed, reproduced, and verified. Thanks — the diagnosis is right and the fix is in the right place.

Reproduced on `main` first, so the before/after is real:

```
$ chez --script host/chez/cli.ss build -m ffiapp.main -o /tmp/ffi-src
$ /tmp/ffi-src
Unhandled exception: Attempting to call unbound fn: #'jolt.ffi/layout-size
```

and with the patch, `FFI-APP 8 4 4 2.5 true`.

`bld-boot-loaded` is supposed to mean "the namespaces the app image inherits", and that image is exactly `bld-runtime-manifest`, which ends at `loader.ss` / `java/ffi.ss`. Snapshotting at `build-binary` time made it "whatever this process happens to have loaded" instead, which stops being the same set the moment the driver is not a release binary. Taking the baseline where `loaded-ns` is seeded makes the two agree by construction, and `java/ffi.ss` adds nothing to `loaded-ns`, so the boundary lands where it should. #756 closed the release half of this from the other side (`ldr-cli-aot?`); this closes the half `bld-emit-cli-aot` never runs for. `hashtable-copy … #f` gives an immutable table, so nothing can drift it later either.

**One real problem with the regression, now fixed on the branch: it was vacuous whenever `target/dev/flat.so` was fresh.** `make devboot` runs `bld-emit-cli-aot` too — `target/dev/flat.ss` carries `(ldr-mark-cli-aot! "jolt.ffi")` — so `bin/jolt build` takes the cached image and passes with or without the fix. `devbootsmoke` is a CI gate, so that is the state the gate would normally run in. A/B against the pre-fix sources with a fresh cache:

| | result |
|---|---|
| `bin/jolt build` (cache used) | `FFI-APP 8 4 4 2.5 true` — passes with the bug present |
| `JOLT_NO_DEVCACHE=1 bin/jolt build` | `Attempting to call unbound fn: #'jolt.ffi/layout-size` |

`bin/jolt` takes `JOLT_NO_DEVCACHE` now and the check sets it.

Other follow-ups pushed:

- Folded the check into build-smoke.sh's existing `[ "$jolt" != "bin/jolt" ]` block. It was an unconditional second source-mode build mid-file, so under `JOLT_BIN=bin/jolt` it ran the same build twice. It reports the value it got on failure now, like its neighbours.
- Kept the `export!` timing note in `libadd/core.clj` — it explains why `jolt_lookup` resolves before `jolt_library_init` returns, which is unrelated to this bug — alongside the new one.
- Comments on `bld-boot-loaded` and `bld-require-closure` naming the manifest tie and how `ldr-cli-aot?` relates now that the two agree.
- Changelog entry.

Note on the library half: `build-lib-smoke.sh` runs under `JOLT_BIN=target/release/jolt` in the gate, where the release binary was already correct — so the new `point_size` export is ABI coverage (a C caller reaching stdlib Clojure through `jolt_lookup`) rather than a gate on this bug. Worth having; the executable case is what gates the fix.

Local gates on the merged branch, macOS arm64 / Chez 10.4.1: `make test` green (ci gate, 90 targets), and `buildsmoke buildlibsmoke devbootsmoke makefilesmoke` re-run green on the final tree.

PR #794, opened 2026-08-30
A File's path is normalized, the way every JVM constructor normalizes it

### LINKED ISSUE #793: java.io.File and clojure.java.io/file do not normalize duplicate or trailing path separators
The JVM's one-arg `File` constructor runs its path through `FileSystem.normalize()`, which collapses duplicate separators and drops a trailing one. jolt keeps the string as given. The same gap shows up one layer up in `clojure.java.io/file`, whose multi-arg form is documented as equivalent to `(File. parent child)` but in jolt does not normalize the join seam.

Scope is exactly duplicate and trailing separators. `.` and `..` are not resolved by the JVM constructor either, and jolt matches there.

Filing this as a question as much as a report: if keeping the path verbatim is deliberate, I would rather understand the reasoning. My read of the source is that it is not, and I have set out why below.

## Measured

jolt `v0.7.29-10-g09af3997` against Clojure 1.12.3 on OpenJDK 25. The JVM side was run from an isolated directory with an explicit `deps.edn` so no ambient dependency could affect the result.

| expression | JVM | jolt |
|---|---|---|
| `(File. "/a/b//c")` | `/a/b/c` | `/a/b//c` |
| `(File. "/a/b/")` | `/a/b` | `/a/b/` |
| `(File. "/a//b//c")` | `/a/b/c` | `/a//b//c` |
| `(File. "//a/b")` | `/a/b` | `//a/b` |
| `(File. "a//b")` | `a/b` | `a//b` |
| `(File. "/a/./b")` | `/a/./b` | `/a/./b` (agrees) |
| `(File. "/a/b/../c")` | `/a/b/../c` | `/a/b/../c` (agrees) |
| `(File. "/a/b/" "c")` | `/a/b/c` | `/a/b/c` (agrees) |
| `(File. "/a//b" "c")` | `/a/b/c` | `/a//b/c` |
| `(io/file "/a/b/" "c")` | `/a/b/c` | `/a/b//c` |
| `(io/file "/a/b/" "c" "d")` | `/a/b/c/d` | `/a/b//c/d` |

## Three distinct paths through `host/chez/java/io.ss`

**The one-arg constructor.** `jolt-make-file` (io.ss:216) stores the path verbatim and joins any extra segments with a plain `string-append`. There is no normalization step.

**The two-arg constructor.** `jolt-file-join` (io.ss:1434) strips a trailing separator from the parent and leading separators from the child, so `(File. "/a/b/" "c")` is correct. It normalizes the join *seam* only, though, and never looks inside either argument. That is why `(File. "/a//b" "c")` still yields `/a//b/c`.

**`clojure.java.io/file` reaches neither.** io.ss:1110 registers it directly:

```scheme
(def-var! "clojure.java.io" "file" jolt-make-file)
```

So the multi-arg form goes through `jolt-make-file`'s `string-append` loop rather than `jolt-file-join`. Clojure's own docstring for `io/file` says the multiple-arg versions are equivalent to `(File. ^File parent ^String child)`, which normalizes. This is probably the higher-impact half: `(io/file dir name)` is common, and a directory string read from configuration or an environment variable often carries a trailing separator.

## Why this reads as an oversight rather than a decision

1. `jolt-make-file`'s own comment (io.ss:213) names JVM parity as the standard: "The File keeps the path AS GIVEN (like the JVM: new File(\"rel\").getPath() is \"rel\")". That comment is about a different property, a relative path not being made absolute, which is genuinely JVM-correct. But the fidelity target is stated outright.
2. `jolt-file-join`'s comment (io.ss:1433) says all four seam cases were "checked against the JVM". The same standard is being applied deliberately one function away.
3. `jolt-file-join` arrived in 3e68b9cb (#528), "Fix the host gaps found running libraries' own test harnesses", a batch of eleven reactive parity fixes. Its comment names the trigger: ring's resource middleware building `"assets/"` plus `"index.html"`. So the normalization that exists is there because a library hit a doubled separator, not because a normalization boundary was drawn.
4. No comment, document, or commit message states a policy of keeping the path verbatim. The only two mentions of normalization in `io.ss` are the two quoted above.

## A practical route in

`file-create-temp` (io.ss:1455) builds its path as `(string-append d "/" prefix ...)` where `d` is `$TMPDIR`. On macOS that value ends in a separator, so `File/createTempFile` returns `.../T//name` where the JVM returns `.../T/name`. Routing that construction through `jolt-file-join` would sidestep the symptom even if the constructor stays as is.

## Relevant to whichever fix you choose

Pointing `clojure.java.io/file` at `jolt-file-join` fixes the doubled separator, but it would also quietly accept a case the JVM rejects. Clojure's `io/file` puts the child through `as-relative-path`, which throws on an absolute one. The two-arg constructor has no such contract, and jolt already agrees with the JVM there:

| expression | JVM | jolt |
|---|---|---|
| `(io/file "/a/b" "/c")` | throws `IllegalArgumentException: /c is not a relative path` | `/a/b//c` |
| `(File. "/a/b" "/c")` | `/a/b/c` | `/a/b/c` (agrees) |

So what is missing is specifically Clojure's `as-relative-path` contract, not anything about the join itself. That places the fix at the `io/file` registration rather than inside `jolt-file-join`. It also means the obvious repair would convert a JVM error case into a silent success, because `jolt-file-join` strips leading separators from the child. Worth deciding on deliberately rather than inheriting it from the fix.

## Adjacent, and not part of this bug

`jolt-list-dir`'s comment (io.ss:238) says "A trailing slash is dropped the way the File constructor normalizes it away." The constructor does not do that. The function is still correct, because it does its own trailing-separator trim inline at the `trimmed` binding (io.ss:243). So this is a stale explanation attached to working code, not a second defect. Mentioning it only because it is load-bearing reasoning sitting next to the affected function.

## Verification

Verified against `upstream/main` at `09af3997` (2026-08-31), confirmed current with `git ls-remote`, working tree clean against that ref. `host/chez/java/io.ss` is untouched between `0e7fe7fc` and `09af3997`, and the measurements were re-run at the later ref rather than carried across.

Searched the tracker for duplicates with several differently-shaped queries, including `File normalize separator`, `duplicate slash path`, `getPath`, `File constructor`, `trailing slash`, `createTempFile` and `file-join`. All empty, positive-controlled against a query that does return hits.

burinc:
Fixes #793.

Two commits, and the second is deliberately separable — see "The second commit is a decision, not a fix" below.

## What was wrong

Every JVM `File` constructor runs its path through `FileSystem.normalize()`, so a `File`'s path is always normalized: runs of `/` collapse to one, a trailing `/` is dropped. jolt kept the string as given.

```
                           JVM        jolt (before)
(File. "/a/b//c")          /a/b/c     /a/b//c
(File. "/a/b/")            /a/b       /a/b/
(File. "/a//b" "c")        /a/b/c     /a//b/c
(io/file "/a/b/" "c")      /a/b/c     /a/b//c
(io/file "/a/b/" "c" "d")  /a/b/c/d   /a/b//c/d
(io/as-file "/a//b/")      /a/b       /a//b/
```

`.` and `..` are not resolved by the JVM constructor either, and jolt already matched there. That is `getCanonicalPath`, which `canonical-path-test.clj` covers.

The visible route in is `createTempFile`: `$TMPDIR` ends in `/` on macOS, so every temp file came back carrying a doubled separator.

## Where the fix went

Not in `jolt-file-join`, and not at the `io/file` registration. Both were tempting and both are wrong places.

`jolt-file-join` already normalized, but only the **join seam** — it strips a trailing separator from the parent and leading ones from the child, and never looks inside either. That is why `(File. "/a//b" "c")` still gave `/a//b/c`.

The deeper problem is that there are nine `make-jfile` call sites and only one is the constructor entry point. `as-file`, the `file:` URL coercion, `createTempFile`, `getParentFile` and `listRoots` all build a `jfile` directly, which is why `io/as-file` had the bug too. So normalization goes in the record's own protocol, where the JVM's actual invariant lives and where the next call site added cannot bypass it.

## The second commit is a decision, not a fix

**Drop the second commit if you would rather `io/file` kept joining an absolute child.** It changes behavior on its own and it is your call, not mine. Each commit carries its own `CHANGELOG.md` entry for exactly this reason, so dropping one leaves the other's entry intact and correct.

But please do not take the first commit alone. `clojure.java.io/file` is not the `File` constructor: Clojure puts every child through `as-relative-path`, which throws on an absolute one, while the two-arg constructor joins it.

```
(io/file "/a/b" "/c")   JVM raises IllegalArgumentException "/c is not a relative path"
(File. "/a/b" "/c")     JVM answers /a/b/c
```

Normalization alone **hides** this rather than fixing it. `"/a/b"` joined to `"/c"` produces `/a/b//c`, which now collapses to `/a/b/c` — a plausible-looking answer to a call the JVM refuses. Shipping the first commit by itself trades a visibly wrong path for a silently wrong one, and the difference matters in practice: `/a/b//c` is something a person notices in a log, whereas `/a/b/c` quietly pointing at the wrong file is not.

Only `io/file` gains the contract. The constructor keeps joining, and the gate asserts both side by side rather than one of them.

On the risk of taking it: a call site that newly throws under this commit is a call site that is **already broken for JVM Clojure**, because `io/file` has always rejected an absolute child there. So the change cannot break code that works on the JVM today. It can only surface code that jolt was quietly accepting and the JVM was not, which is the same class of thing the corpus and `certify` exist to catch.

## Gate

New: `test/chez/path-normalize-test.clj`, wired into `host/chez/smoke.sh` next to `canonical-path-test.clj`, same self-check-plus-marker shape.

Every expected value is a JVM value, measured against Clojure 1.12.3 on OpenJDK 25 rather than reasoned about — including the cases where normalization must **not** fire (`/` stays `/`, `""` stays `""`, `/a/./b` and `/a/b/../c` unchanged, `..//` becomes `..`), and the `as-relative-path` throw across arities.

## What I ran

| gate | result |
|---|---|
| `make smoke` | 192 passed, 0 failed (up one, the new gate) |
| `make unit` | green |
| `make corpus` | 4759/4769, **0 NEW divergences**, 10 known |
| `make certify` | 108/108 divergences known, **0 new, 0 stale** |
| `make test` | **OK: ci gate passed (90 targets)**, test gate passed (3 targets) |
| `make libconformance` | **not run** — no library checkout on this machine |

`libconformance` is a genuine gap rather than a pass, so the 47-suite replay is unverified from my side. Since this touches path handling that libraries lean on, that is the coverage I would most want before merge.

`make remint` was not run: `host/chez/java/io.ss` is a runtime host file loaded by `rt.ss`, not a compiler source, so the seed is unaffected. Please tell me if that reading is wrong.

## On the underlying question

#793 asked whether the verbatim path was deliberate, and I have written this PR on the assumption it was not — the reasoning is in the issue. If it was in fact a decision, this can be closed and I would still rather know the reason than be right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,16 @@ stays authoritative.
   `.` and `..` are still not resolved. The JVM constructor does not resolve them
   either; that is `getCanonicalPath`.
 
+- **`clojure.java.io/file` joined an absolute child instead of rejecting it.**
+  `io/file` is not the `File` constructor: Clojure puts every child through
+  `as-relative-path`, so `(io/file "/a/b" "/c")` raises `IllegalArgumentException`
+  while `(File. "/a/b" "/c")` answers `/a/b/c`. jolt joined it either way.
+
+  Worth knowing that normalization alone would have hidden this rather than
+  fixed it — `/a/b` joined to `/c` gives `/a/b//c`, which collapses to a
+  plausible-looking `/a/b/c` answer to a call the JVM refuses. A call site that
+  newly raises here was already broken for JVM Clojure.
+
 - **`jolt run` could not write any closure `clojure.core` makes.** `cycle`,
   `repeat`, `partial`, `comp` and the rest refused with "captured local … was
   optimized into the compiled code" — while a default `jolt build` wrote them
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,14 @@ stays authoritative.
   `.` and `..` are still not resolved. The JVM constructor does not resolve them
   either; that is `getCanonicalPath`.
 
+  The two-arg constructor is `resolve(normalize(parent), normalize(child))` — it
+  normalizes each argument, and resolve is not a plain join. An empty parent
+  resolves against `getDefaultParent()`, so `new File("", "")` is `/`, not `""`,
+  which jolt got wrong in both directions before. (If you go measuring this
+  yourself: resolve grew its `child == "/"` case in JDK 21, so through JDK 20
+  `new File("/a/b", "/")` answered `/a/b/`, a path with a trailing separator no
+  one-arg constructor can produce. 21 onward it is `/a/b`.)
+
 - **`clojure.java.io/file` joined an absolute child instead of rejecting it.**
   `io/file` is not the `File` constructor: Clojure puts every child through
   `as-relative-path`, so `(io/file "/a/b" "/c")` raises `IllegalArgumentException`
@@ -63,6 +71,19 @@ stays authoritative.
   JVM, missing here entirely). And `io/make-parents` builds `(apply io/file f
   more)` on the JVM, so it carries the same contract: an absolute child raises
   instead of quietly joining.
+
+- **`(io/file nil)` and `(io/as-file nil)` answered an empty `File`, not `nil`.**
+  Clojure extends `Coercions` to `nil`, so both are `nil` on the JVM. A `File`
+  whose path is `""` is not a harmless stand-in for that: it names the process's
+  working directory, so a nil that should have blown up at the coercion instead
+  went on to read or write the wrong file. `as-relative-path` goes through
+  `as-file`, so a nil child now raises there too, the way `(io/file "/a" nil)`
+  does on the JVM, rather than joining as an empty segment.
+
+  The `File` constructor null-checks for the same reason: `(File. nil)` and
+  `(File. "/a" nil)` raise `NullPointerException` instead of reading the nil as
+  `""`. `(File. nil "c")` still answers `c` — a null *parent* is the one the
+  two-arg constructor accepts, and it means the child alone.
 
 - **`jolt run` could not write any closure `clojure.core` makes.** `cycle`,
   `repeat`, `partial`, `comp` and the rest refused with "captured local … was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,14 @@ stays authoritative.
   plausible-looking `/a/b/c` answer to a call the JVM refuses. A call site that
   newly raises here was already broken for JVM Clojure.
 
+  Review follow-ups, all JVM-measured: `as-relative-path` goes through `as-file`
+  first, so the thrown message names the normalized path — `(io/file "/a/b"
+  "//c")` says `/c is not a relative path`, not `//c`. `io/as-relative-path`
+  itself is now registered as a var (public API in `clojure.java.io` on the
+  JVM, missing here entirely). And `io/make-parents` builds `(apply io/file f
+  more)` on the JVM, so it carries the same contract: an absolute child raises
+  instead of quietly joining.
+
 - **`jolt run` could not write any closure `clojure.core` makes.** `cycle`,
   `repeat`, `partial`, `comp` and the rest refused with "captured local … was
   optimized into the compiled code" — while a default `jolt build` wrote them
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,32 @@ stays authoritative.
 
 ### Fixed
 
+- **A `File`'s path was kept exactly as given, not normalized.** Every JVM `File`
+  constructor runs its path through `FileSystem.normalize()`, so runs of `/`
+  collapse to one and a trailing `/` is dropped: `new File("/a/b//c").getPath()`
+  is `/a/b/c`. jolt answered `/a/b//c`. The visible route in was
+  `createTempFile`, since `$TMPDIR` ends in `/` on macOS and every temp file came
+  back carrying a doubled separator.
+
+  `clojure.java.io/file` had the same gap and a second one under it. It is
+  registered as the bare file constructor, whose multi-arg loop appends `/`
+  unconditionally, so `(io/file "/a/b/" "c")` gave `/a/b//c` where the JVM gives
+  `/a/b/c` — Clojure's own docstring says the multi-arg versions are equivalent
+  to `(File. parent child)`. That is the half more likely to bite, since
+  `(io/file dir name)` is everywhere and a directory read from config or the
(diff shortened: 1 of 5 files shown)


yogthos:
Merging. Reviewed, and I ran the gate you could not: `make libconformance` is **50 libraries, 47 ok, 0 better, 0 regressed, 3 skipped**. Your reading on `make remint` was right — `host/chez/java/io.ss` is runtime, not compiler source, and the self-host fixpoint holds.

Follow-ups pushed as one commit on top:

- The two-arg constructor is `resolve(normalize(parent), normalize(child))` — it normalizes each *argument*, and resolve is not a plain join. An empty parent resolves against `getDefaultParent()`, so `new File("", "")` is `/` where jolt said `""`. `jolt-file-join` spells resolve out now.
- `Coercions` is extended to `nil`, so `(io/file nil)` and `(io/as-file nil)` are `nil`, not a `File` whose path is `""` — which names the cwd. Same silently-wrong-file shape the absolute-child fix is about, one seam over. `as-relative-path` and `make-parents` inherit it, and the constructor null-checks its own arguments: `(File. nil)` and `(File. "/a" nil)` raise.
- `jolt-path-normalize` scans before it copies, so an already-normal path comes back as-is with nothing allocated. A jfile is built per entry on every directory listing.

One note on measurement, since it cost me an hour. Three of your separator-only-child rows I initially "corrected" — `(File. "/a/b" "/")` and friends — because a real JDK 20 answers `/a/b/`. You were right: `UnixFileSystem.resolve` only grew its `child == "/"` case in JDK 21. Checked on 20, 21 and 26; jolt matches 21+. Both files now say so, so the next person measuring on an old JVM does not "fix" it back the way I nearly did.

To settle your last question: yes, the verbatim path was not deliberate.

The whole gate now runs green on a real JVM (Clojure 1.12.5, OpenJDK 26) as well as on jolt, so every expectation in it is a measured JVM value rather than a claim about one. Thanks — the reasoning about *where* the invariant belongs was the valuable part, and it is what made the follow-ups small.

PR #795, opened 2026-08-30
Normalize File paths like the JVM constructor

### LINKED ISSUE #793: java.io.File and clojure.java.io/file do not normalize duplicate or trailing path separators

yogthos:
Fixes #793.

Every `java.io.File` constructor on the JVM runs its path through `FileSystem.normalize`: runs of separators collapse and one trailing separator drops (the root `/` alone keeps its), while `.` and `..` are left alone. jolt stored the path as given. Three changes close the gap:

- **The normalize lives in the jfile record constructor** (`host/chez/java/io.ss`), so "a File's path is always normalized" is an invariant — every creation site answers normalized paths, including `getParentFile`, `listFiles` children, `io/as-file`, and URL coercion, which no site can opt out of.
- **`jolt-file-join` now implements `UnixFileSystem.resolve(normalize(parent), normalize(child))`** instead of only trimming the join seam, so `(File. "/a//b" "c")` is `/a/b/c`. The two-arg ctor stores the resolved result verbatim (a separator-only child yields `/parent/`, JVM-faithful). `File/createTempFile` joins through it, so `$TMPDIR` ending in `/` no longer produces `.../T//name`.
- **`clojure.java.io/file`'s multi-arg forms carry Clojure's `as-relative-path` contract** (registered as `io/as-relative-path` too): an absolute child throws `IllegalArgumentException` with the JVM's `… is not a relative path` message rather than silently joining — the gap the issue called out as worth deciding deliberately. `io/make-parents` routes through the same semantics, since the JVM builds it from `(apply file f more)`.

Scope is exactly the issue's: duplicate and trailing separators. `.`/`..` resolution is untouched, matching the JVM.

New gate `test/chez/file-normalize-test.clj` (wired into `smoke.sh` as the `FILE-NORMALIZE` case) covers the issue's measured table plus the as-relative-path throws, `createTempFile`, and `listFiles` children.

Verified: `make smoke` 192/192, `make unit` 1532/1532, `make corpus` 0 new divergences (10 known allowlisted), `make documented` 35/35, `make ioscaling` and `make grenadine` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+`java.io.File` normalizes its path the way the JVM constructor does: runs of
+separators collapse and one trailing separator drops (the root `/` alone keeps
+its), while `.` and `..` are still left alone (#793). The normalization lives
+in the jfile record constructor, so every File creation site — including
+`getParentFile`, `listFiles` children, `io/as-file`, and `File/createTempFile`,
+whose temp paths no longer carry a doubled separator when `$TMPDIR` ends in `/`
+— answers normalized paths. The two-arg constructor now implements
+`UnixFileSystem.resolve` over normalized inputs, and `clojure.java.io/file`'s
+multi-arg forms gained Clojure's `as-relative-path` contract (registered as
+`io/as-relative-path` too): an absolute child throws
+IllegalArgumentException (`… is not a relative path`) instead of quietly
+joining, matching the JVM. `io/make-parents` follows `io/file` semantics for
+the same reason.
+
 The build keeps one toolchain end to end now. When make provisions the pinned
 Chez + xPack GCC, the standalone-binary link runs through that same provisioned
 GCC — `JOLT_CC`, honored by `build.ss`'s `bld-cc` — instead of a bare `cc`,
--- a/host/chez/java/io-streams.ss
+++ b/host/chez/java/io-streams.ss
@@ -867,7 +867,9 @@
             ((char=? (string-ref p i) #\/) (mkdirs! (substring p 0 i)))
             (else (loop (- i 1)))))))
 (define (apply-make-file-path args)
-  (jfile-path (apply jolt-make-file args)))
+  ;; the JVM's make-parents is (.mkdirs (as-file (apply file f more))) — io/file
+  ;; semantics, not the bare ctor, so an absolute child throws here too
+  (jfile-path (apply jolt-io-file args)))
 (def-var! "clojure.java.io" "make-parents" jio-make-parents)
 
 ;; io/delete-file: delete the file; raise unless :silently truthy.
--- a/host/chez/java/io.ss
+++ b/host/chez/java/io.ss
@@ -13,8 +13,28 @@
 ;; natives-meta.ss / records.ss / printing.ss (jolt-type / instance-check /
 ;; jolt-str-render-one, which it extends).
 
-(define-record-type jfile (fields path) (nongenerative jolt-jfile-v1))
+(define-record-type (jfile %make-jfile jfile?) (fields path) (nongenerative jolt-jfile-v1))
 (define (jolt-file? x) (jfile? x))
+
+;; java.io.FileSystem#normalize (Unix), which every JVM File constructor runs
+;; its path through: a run of separators collapses to one, and one trailing
+;; separator drops — except on the root "/" alone. "." and ".." are NOT
+;; resolved, and a relative path stays relative. Routing it through the record
+;; constructor makes "a File's path is always normalized" an invariant here
+;; too: no creation site below can store a verbatim path, same as the JVM.
+(define (jolt-normalize-path p)
+  (let ((n (string-length p)))
+    (let ((buf (make-string n)))
+      (let collapse ((i 0) (j 0))
+        (if (= i n)
+            (let ((j (if (and (> j 1) (char=? (string-ref buf (- j 1)) #\/)) (- j 1) j)))
+              (if (= j n) p (substring buf 0 j)))
+            (let ((c (string-ref p i)))
+              (if (and (char=? c #\/) (> j 0) (char=? (string-ref buf (- j 1)) #\/))
+                  (collapse (+ i 1) j)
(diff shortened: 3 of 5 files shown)


yogthos:
Closing in favor of #794, which covers the same ground (record-level normalize, resolve-style join, io/file contract). Follow-up fixes for the remaining JVM gaps — as-relative-path's normalized message, the missing io/as-relative-path var, and io/make-parents routing — are pushed there as be9ca0ea.

PR #801, opened 2026-08-31
Fix unsound unsafe-fixnum hash ops on narrow Chez targets (tpb32l support)

jasalt:
Patch proposal for issue that came up while experimenting with WASM support through Petite Chez `tpb32l` target which I also found to have been used for Racket Wasm REPL recently https://racket.discourse.group/t/web-based-racket-repl-running-on-wasm/4277.

Related downstream experimentation: https://github.com/jasalt/jolt-wasm-raylib-experiment/tree/main/experiments/EXP-014-jolt-tpb32l-emscripten

Pi coding agent & gpt-5.6-sol generated solution, looks reasonable to me but it is beyond my understanding of the low level details in play here.
_____
LLM generated content begin

## Problem

Jolt's Chez host layer (`host/chez/hasheq.ss`, `host/chez/collections.ss`) uses `#3%` unsafe fixnum primitives for the Murmur3/hash-combine engine and the HAMT bitmap arithmetic. The soundness comment rests on "all unsigned 32-bit values are fixnums," which holds only on wide-fixnum (64-bit) Chez targets. On 32-bit targets such as `tpb32l`, the positive fixnum ceiling is 2^29-1, so valid Java `int` values — HAMT hashes with bit 30 set, folded 32-bit intermediates — are bignums, and the unsafe `#3%fx` path is unsound.

## Minimal example

Before the fix, e.g. `collections.ss` masked hashes with a raw unsafe ops:

```scheme
(define (chunk h shift) (fxand (fxsra h shift) 31))
(define (bitpos h shift) (fxsll 1 (chunk h shift)))
```

After the fix, a foldable width-selecting macro picks the fast or generic path at macro expansion time, and call sites go through it:

```scheme
(define-syntax hamt-and
  (syntax-rules () ((_ a b) (if (fixnum? hmask) (fxand a b) (bitwise-and a b)))))

(define (chunk h shift) (hamt-and (hamt-sra h shift) 31))
(define (bitpos h shift) (hamt-sll 1 (chunk h shift)))
```

On a 64-bit target the literal `(fixnum? #xFFFFFFFF)` test folds to true, so generated code retains the existing `fx` chain with no runtime dispatch; on `tpb32l` it folds to false and generic exact-integer ops are selected. `host/chez/hasheq.ss` does the same for its `#3%fx*` primitives via a `hash-fx*` macro family.

## Change

- `host/chez/hasheq.ss`: new `hash-fx*` macro family (`hash-fx+ - * and ior xor sll sra srl = < >=`); every `#3%fx` call site in the hash engine routed through them; soundness comments rewritten to describe the width-conditional selection instead of a 61-bit-ceiling invariant.
- `host/chez/collections.ss`: new `hamt-and ior not sll sra popcount` macros; HAMT chunk/bitpos/bitmap arithmetic and `key-hash` masking routed through them.
- `test/chez/tpb32l-hash-test.ss` (new): tpb32l-only regression asserting JVM-identical Murmur3 and 32-bit wrap results.

## Artifacts

- Patch source: https://github.com/jasalt/jolt-wasm-raylib-experiment/tree/main/experiments/EXP-014-jolt-tpb32l-emscripten/patches
- Regression: `test/chez/tpb32l-hash-test.ss`


yogthos:
Reviewed and gated. The premise holds: on a 32-bit Chez the positive fixnum ceiling is 2^29-1, so an ordinary Java-int hash is a bignum and `#3%fx` on it answers nonsense rather than raising. Worth noting this repo had already reached the same conclusion on the other host — `host/gambit/prelude-shims.ss` maps its `fx` aliases to generic arithmetic with the comment *"Gambit-js fixnums are ~30-bit (tagged), so 32-bit hash values overflow"*. Same problem, same shape of fix.

I pushed a follow-up commit rather than asking for changes, since three of the four points needed measurement to settle.

**The `collections.ss` half did not fold.** `hasheq.ss` tested the literal `(fixnum? #xFFFFFFFF)`, which cp0 folds away. `collections.ss` tested `(fixnum? hmask)` — a top-level variable, which cp0 cannot constant-propagate, because a top-level binding can be redefined. `chunk`, on the HAMT's hottest path, expanded to:

```scheme
(if (#3%fixnum? hmask)
    (#3%fxand 31 (if (#3%fixnum? hmask) (#3%fxsra h shift) (#3%bitwise-... h shift)))
    (let ([g0 (if (#3%fixnum? hmask) (#3%fxsra h shift) ...)]) ...))
```

Four runtime tests against a global, with the shift arm duplicated. A few call sites had also moved to unconditional generic `=` and `-`, which costs 64-bit builds nothing in return. A/B/A against main (5 runs each, order A→B→A): **hash-eq 1.04x, keyed-lookup 1.02x** slower. Under the repo's 1.1x ceiling, but avoidable.

**`key-hash` lost a fallback that wasn't about width.** main had `(if (fixnum? h) (fxand h hmask) (bitwise-and h hmask))` — the comment above it says why: an extension type can return a bignum from `equal-hash` on *any* machine. Folding that into the width macro dropped it on 64-bit. It's back, on both arms.

**What the follow-up does.** The 18 hand-written `(if (fixnum? ...) fast generic)` wrappers collapse into one `define-width-op` in `hasheq.ss` that selects the operator at *expand* time via `syntax-case`, so exactly one arm reaches the compiler — no runtime test, no duplicated arms, and no dependence on cp0 folding anything. `collections.ss` declares its HAMT operators through the same form, keeping the safe `fx` spellings it has always used (the bitmap arithmetic is not the measured `#3%` inner chain, and a checked op there fails loudly instead of corrupting a node). `host/gambit/prelude-shims.ss` defines the form to always take the wide arm, since that target's `fx` aliases are already generic for the same reason.

Verified the wide path generates the identical `#3%fx` chain at optimize-level 0 **and** 3. A/B/A of the follow-up vs main: `collections` 19.4/19.6, `hash-eq` 5204/5217, `keyed-lookup` 9382/9473, `transients` 75.8/76.0 ms — parity.

**The narrow arm is testable now.** `test/chez/tpb32l-hash-test.ss` called `(error ...)` unless `(machine-type)` was `tpb32l`, so on every other machine it failed rather than skipped — nothing could run it, and its values duplicated goldens already pinned in `hasheq-test.ss`. Replaced with `JOLT_NARROW_HASH=1`, read at expand time, which forces the narrow arm on ordinary hardware. `make narrowhash` replays the JVM-pinned hash goldens, the value-model suite and the transient HAMT suite through the generic operators, and it's wired into `CI-GATES`. Confirmed it isn't vacuous:

| probe | wide arm | narrow arm |
|---|---|---|
| `(hash-fx+ (expt 2 70) 1)` | `#(536870912 ...)` (garbage) | `1180591620717411303425` |
| `(hamt-and (expt 2 70) 255)` | raises | `0` |

`gate-boot.ss` skips the precompiled `gate.so` when the knob is set — the image's freshness is content-hashed over file bytes, which cannot see an expand-time environment variable, so it would otherwise have quietly tested the wide path.

One scoping note for anyone finding this later: `hasheq.ss` is the only file in the repo that uses `#3%fx` at all, so this closes the *unsound* cases exactly. It is not full `tpb32l` support — other host code still applies checked `fx` ops to int-width values, which raise rather than corrupt.

`make test` is green. The branch also carries one unrelated commit: #802's new arena suite hardcoded `libm.so.6` in the `load-library` ordering row, which does not exist on macOS, so `make test` was red on any Mac while CI stayed green on Linux. That row now asks `load-system-library "m"` for this host's libm.

Thanks for the patch and for the measurements in the linked experiment — the diagnosis was right, and the pb/WASM target is a good thing to be pulling on.

PR #820, opened 2026-09-01
Executors/newCachedThreadPool grows on demand, as on the JVM (#819)

### LINKED ISSUE #819: Executors/newCachedThreadPool is is a fixed 32-worker pool rather than a growing one as on the JVM
Update the pool to follow JVM semantics and grow on demand.

yogthos:
Closes #819.

`Executors/newCachedThreadPool` was a fixed pool of 32 workers, so a burst wider
than 32 concurrent tasks queued behind the ones already running — invisible while
tasks are short, and a deadlock when they are not: 32 tasks blocking on the 33rd
never let it start, with no error and nothing in the code to suggest a ceiling.

The executor now has `ThreadPoolExecutor`'s shape — core workers, a max to grow to,
and a keep-alive after which an above-core worker retires. Growth is decided under
the queue mutex as `queue-depth > idle-workers`, which is the `SynchronousQueue`
handoff the JVM expresses through its queue type; the exit decision and the
live-count decrement happen under one hold, so an enqueue can never read a worker
as live after it has stopped taking work.

- **cached / virtual-thread-per-task**: `(0, Integer.MAX_VALUE, 60s)`, the JVM's own
  numbers. Nothing forked until a task arrives.
- **fixed / single-thread**: unchanged — core = max, eager, never retired (the
  single-thread pool's submission order depends on exactly one worker, forever).
- **work-stealing**: still fixed, a `ForkJoinPool` being core-count-bounded there too.
- **`ThreadPoolExecutor.`**: core, max and keepAlive all reach the pool now; the ctor
  used to size a fixed pool by max and discard the keep-alive.
- **new**: `.getPoolSize`, `.getActiveCount`, `.getCorePoolSize`, `.getMaximumPoolSize`,
  so the growth and the retirement are observable rather than inferred.

## Latent issues this turned up, and fixed

**An enqueue woke every worker, not one.** Workers and `awaitTermination` shared one
condition, so each task broadcast to all waiters: with 130 idle workers, one task woke
130, and 129 took the mutex, found nothing and re-parked. It also fed the growth rule,
so a 200k burst reached 134 threads and 131MB. Task and termination now have a
condition each and an enqueue signals exactly one parked worker (`jolt-cv-signal-one!`,
with its two preconditions written into `locks.ss`). The broadcast predates this branch,
so this is a 3x improvement on 0.8.0 as well as a 17x one on the growing pool.

**The three spellings of shutdown did not do what they say** — each verified against
reference JVM Clojure on Java 21 before changing anything:

- a `submit`/`execute` after shutdown was queued and abandoned (its `.get` waits
  forever); now `RejectedExecutionException`, decided under the queue mutex so a task
  that beats the flag still runs;
- `shutdownNow` answered an empty vector and ran the queue anyway — the method that
  exists to say "do not run the rest" ran the rest; now drops the queued tasks and
  returns them (callable, like the JVM's Runnables). Interrupting *running* tasks is
  still missing and is marked in the source;
- `close` returned the moment the flag was set, so `(with-open [ex …] …)` left its
  tasks running; now blocks until terminated, as the JVM's has since 19.

A worker that dies for any reason other than its task throwing also no longer leaks
its slot, which would have hung `awaitTermination` on a finished pool.

## Measured

200k no-op tasks through `.execute`, same box, against reference Clojure on Java 21:

| | jolt | JVM |
|---|---|---|
| cached pool | 8.9 µs/task (peak 7 threads) | 5.2 µs (peak 8) |
| 4 producers, one pool | 11 µs/task | 2.2 µs |
| single-thread pool | 4.3 µs | 9.1 µs |
| submit + get round trip | 12 µs | 7.2 µs |
| 256 threads created | 21 ms | 22 ms |

and for the pool itself: 152 µs/task and 134 threads before the wake fix, 26.5 µs and
32 threads on 0.8.0. What is left between jolt and the JVM is one mutex and one
park/unpark per handoff where a `SynchronousQueue` transfers without either.

A `bench/executors` row is added so this is watched: nothing in the suite measured
dispatch, and the 17x regression above passed every correctness gate.

## Tests

Twelve new unit rows (`cached-pool`, `executor-shutdown`); nine of them render exactly
what reference JVM Clojure renders for the same expression. The rest are jolt's own
(a fiber reaching the growth path; the `ThreadPoolExecutor` arity the JVM ctor does not
have; the eager core threads).

Gates: `unit`, `flow`, `smoke` (192/0), `corpus` (0 new divergences), `documented`,
`threadsafety`, `fibers`, `gosm`, `asynctimer`, `interruptnest`, `sci`, `cts` (matches
baseline), `lockcheck`, `parkcheck`.

## Known follow-ups

- `ArrayBlockingQueue` put/take is 8.5 µs against the JVM's 0.39 µs. The cost is
  `jolt-cv-wake!` taking a global mutex and hashtable on every operation to check for
  parked fibers; a fast path there touches every wake in the runtime and wants its own
  issue.
- `shutdownNow` cannot interrupt running tasks — the workers carry no interrupt flag.
- Multi-producer contention needs queue striping or per-worker handoff slots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`Executors/newCachedThreadPool` grows on demand (#819).** It was a fixed pool
+  of 32 workers. A burst of more than 32 concurrent tasks queued behind the ones
+  already running — invisible while tasks are short, and a deadlock when they are
+  not: 32 tasks that block waiting for the 33rd (a fan-out whose children hand
+  their results back through a channel, say) never let it start, with no error and
+  nothing in the code to suggest a ceiling. The pool now has the JVM's shape —
+  core 0, max `Integer/MAX_VALUE`, 60s keep-alive — so nothing is forked until a
+  task arrives, a task that no idle worker is waiting to take starts one, and a
+  worker retires after 60 seconds idle.
+
+  It is still a POOL, which is what keeps that unbounded maximum from meaning a
+  thread per task: 2000 trivial tasks submitted back to back peak at **4** workers
+  on an idle 8-core box (5000 of them, 3), because the handful that keep up with
+  them are idle again by the time the next one lands. How far it grows is a
+  property of how far the producer runs ahead of the workers — the same 2000-task
+  burst pinned to a single core peaks at 136 — and of the tasks themselves: 1000
+  tasks that each sleep 50ms do reach 558 workers. Which is the JVM's own answer to
+  the same bursts, measured on the same box against reference Clojure: 13 workers
+  for the 2000 trivial tasks, 697 for the 1000 sleepers, and the same 64 for the
+  64 blocking ones. The unbounded growth is concurrency, not churn.
+
+  `newVirtualThreadPerTaskExecutor` maps to the same pool, being equally unbounded
+  on the JVM; a pooled thread stands in for a fresh virtual one, which nothing
+  here can tell apart. `newWorkStealingPool` stays a fixed pool, because a
+  `ForkJoinPool` is sized at `availableProcessors` on the JVM too. `newFixedThreadPool`
+  and `newSingleThreadExecutor` are unchanged: every worker eager, none retired,
+  and for the single-thread pool the submission ORDER that depends on there being
+  exactly one worker forever.
+
+  `ThreadPoolExecutor.`'s `corePoolSize`, `maximumPoolSize` and `keepAliveTime`
+  now all reach the pool it builds — the core workers eager, the rest on demand up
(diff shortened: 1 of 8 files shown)


yogthos:
Correction to the follow-up list above, from measuring it in #821: the `ArrayBlockingQueue` gap is **not** `jolt-cv-wake!`'s global mutex and hashtable. Those are ~150ns of a 2000ns put+take pair. Priced properly (one thread, queue never empty or full, so nothing parks), half the cost was generic interop dispatch — the arm that answers for a host shim is the last one in the chain, so every call walked nine arms first. #821 has the numbers and the fix; the ABQ pair is 2000ns -> 1025ns.

The probe that pointed at the global tables was a producer/consumer loop dominated by park/unpark scheduling: stubbing the tables out of it made it read *slower*. Worth naming, since that is how the wrong cause got into this PR's text.

PR #824, opened 2026-09-02
Publish a rolling nightly prerelease from main

burinc:
A rolling prerelease built from `main` daily, so a downstream project can test against unreleased jolt without building it from source.

## Why

A break lands on `main` and nobody downstream sees it until the next tag. jolt moved 45 commits in about 30 hours between `v0.8.0` and this branch's base — that is a long window in which a regression is only visible to whoever builds from source.

Two recent breaks make the case concretely. `ffi/write`'s argument order moved in #802, and the two spellings are indistinguishable at runtime because an offset and a value are both integers, so an unmigrated call writes to the wrong place and reports nothing. And #813's provider-declaration change means a library that has not declared `:jolt/provides` fails to resolve its classes — where a consuming project's namespace is dropped from a test suite that then exits 0. Both are the kind of thing a downstream suite catches immediately and a release cadence surfaces late.

## What it does

`schedule` at 17:00 UTC builds `main` and publishes it as a **draft prerelease** at the rolling tag `vnightly`, replaced daily.

The tag name is not arbitrary. jolt's own install script does `tag="v${version#v}"` and `filename="jolt-${tag}-${target}.tar.gz"`, so `vnightly` makes `--version nightly` resolve to `jolt-vnightly-<target>.tar.gz` — a nightly is installable with the shipped installer, unchanged.

It publishes as a draft first, exactly as a real release does and for the same reason: a nightly that fails the gates stays a draft nobody can install, and the previous nightly remains the newest thing available.

The tagged-release path is untouched. `publish` and `homebrew` are already guarded to `refs/tags`, so a scheduled run never reaches them.

## `releases/latest` is unaffected — verified, not assumed

The nightly sets `prerelease: true`, and GitHub's `releases/latest` returns the most recent **non-prerelease, non-draft** release. Checked empirically rather than from the docs, against a repo that actually publishes both:

```
neovim/neovim releases:
  nightly   prerelease=true   published=2026-09-02   <- newest
  v0.12.5   prerelease=false  published=2026-08-23

neovim/neovim releases/latest:
  v0.12.5                                            <- the stable one
```

This matters downstream rather than in the abstract: at least one project in this ecosystem installs jolt in CI via a script that resolves `releases/latest`. A nightly that changed what that returns would silently move every such CI job onto an unreleased build. It does not.

## Second commit: making it testable before it merges

Two gaps that made the feature impossible to try.

**It was reachable only from the schedule**, so the only way to exercise it was to merge and wait for cron. `workflow_dispatch` now takes a `nightly` boolean, and a single `JOLT_NIGHTLY` expression feeds every guard so they cannot drift apart.

**The install script hardcoded `jolt-lang/jolt`**, so a nightly built on a fork was unreachable by the very command the release notes tell people to run — `--version nightly` looked in the upstream repo, where a fork's rolling release does not exist. It now honours `JOLT_REPO` and a `--repo` flag, defaulting exactly as before:

```
JOLT_REPO=someone/jolt install --version nightly
```

That is useful beyond forks: it is what lets anyone test a candidate build without it first being published upstream, which is the same reason the nightly exists.

## Notes for review

The cron time, the `vnightly` tag name, and whether a rolling prerelease is wanted at all are the maintainer's calls — happy to change any of them or to drop this if the cadence is not what you want.

Not verified: I have not observed a scheduled run complete end to end, since that requires the workflow to be on a branch the schedule reaches. `workflow_dispatch` in the second commit exists precisely so that can be tried before merging.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,13 @@ on:
   push:
     tags:
       - 'v*'
-  workflow_dispatch: {}   # dry-run the build matrix without tagging
+  workflow_dispatch:
+    inputs:
+      nightly:
+        description: "Build and publish the rolling nightly prerelease, as the schedule does"
+        type: boolean
+        required: false
+        default: false
   # A rolling prerelease built from main daily, so downstream projects can test
   # against unreleased jolt without building it from source. The tagged-release
   # path is untouched: `publish` and `homebrew` are already guarded to
@@ -32,6 +38,10 @@ permissions:
 # `vnightly` makes `--version nightly` resolve to `jolt-vnightly-<target>.tar.gz`,
 # and a nightly is installable with the shipped installer, unchanged.
 env:
+  # Schedule or an explicit dispatch. Without the dispatch half the nightly path
+  # can only ever be exercised by waiting for the cron, which makes it
+  # untestable in a fork and unreviewable in a pull request.
+  JOLT_NIGHTLY: ${{ github.event_name == 'schedule' || inputs.nightly == true }}
   JOLT_TAG: ${{ startsWith(github.ref, 'refs/tags/') && github.ref_name || 'vnightly' }}
 
 jobs:
@@ -308,7 +318,7 @@ jobs:
       # nightly that fails the gates below stays a draft nobody can install, and
       # the previous nightly remains the newest thing available.
       - name: Upload to the nightly prerelease (draft)
-        if: github.event_name == 'schedule'
+        if: env.JOLT_NIGHTLY == 'true'
         uses: softprops/action-gh-release@v2
         with:
           tag_name: vnightly
@@ -677,7 +687,7 @@ jobs:
   publish-nightly:
     name: publish nightly
     needs: [build, libraries, libraries-db, examples, bench]
-    if: github.event_name == 'schedule'
+    if: github.event_name == 'schedule' || inputs.nightly == true
     runs-on: ubuntu-latest
     timeout-minutes: 15
     permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,12 @@
 name: release
 
 # Build the self-contained jolt binary for each platform and attach it to the
-# GitHub Release when a v* tag is pushed. The binary bundles the runtime,
-# compiler, jolt-core + stdlib source, the Chez boots, and a launcher stub, so it
-# runs AND compiles jolt apps with no Chez or cc on the user's machine (jolt-eaj).
+# GitHub Release when a v* tag is pushed — and, daily, publish main as the
+# rolling `vnightly` prerelease (`install --version nightly`) so a downstream
+# project can test against unreleased jolt without building it. The binary
+# bundles the runtime, compiler, jolt-core + stdlib source, the Chez boots, and a
+# launcher stub, so it runs AND compiles jolt apps with no Chez or cc on the
+# user's machine (jolt-eaj).
 #
 # No Apple notarization, mirroring dirge: macOS users who download the tarball
 # clear Gatekeeper quarantine once (`xattr -d com.apple.quarantine jolt`), or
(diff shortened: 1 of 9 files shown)


burinc:
Exercised this via `workflow_dispatch` on a fork, and it immediately found something that makes the case better than the PR description did.

## The nightly caught three first-party libraries broken on current `main`

A dispatch run against `5e0e7a96` failed three library gates, all with the same root cause:

```
library db             No dependency provides Locale — RFC 0014
library db (postgres)  No dependency provides Locale — RFC 0014
library transit-jolt   No dependency provides java.time.format.DateTimeFormatter — RFC 0014
```

`498006c0` (#813) moved host-class provision to per-library declaration, and `jolt-lang/time` has not declared its `:jolt/provides`. Every library that formats or zones a date fails to resolve. Reproduced locally on `v0.8.0-45-g5e0e7a96`:

```
$ jolt -e '(println (java.time.ZoneId/systemDefault))'
Unhandled exception: No dependency provides java.time.ZoneId — ... (RFC 0014)
```

Builds, examples and benchmarks all passed in the same run. Only the library gates caught it.

## Why that is an argument for this PR specifically

The library gates live only in `release.yml`, which triggers on `push: tags: v*` and `workflow_dispatch`. `tests.yml` runs on every push and PR and contains no library gates.

**So library breakage on `main` is invisible until a tag is cut** — at which point the release itself fails its own gates, during a release. This PR's `schedule` trigger is what moves that discovery from release time to the next morning.

That is not a hypothetical benefit argued from first principles. It is the first thing the feature did when run.

## The draft gate also worked, on a failing run

The failing run left `vnightly` as a **draft** with its 8 assets attached and unpublished — exactly the behaviour the PR describes. A nightly that fails the gates stays uninstallable, and whatever was previously newest remains so. Worth noting the failure path is now exercised, not just the happy one.

## What I have not shown

I have not completed an end-to-end install of a published nightly, because the gates fail on current `main` and the draft gate therefore — correctly — refuses to publish one. That test needs either a green `main` or the `jolt-lang/time` declaration landing first.

So: the build, packaging, draft-upload and gate-refusal paths are exercised. The publish and install paths are not yet.

## Separately

The `jolt-lang/time` gap looks like the thing to fix regardless of this PR. Its own `statics!` calls in `src/jolt/time/{zones,zoned,fmt}.clj` name what it should declare, and `host-static.ss`'s comment already says everything naming or formatting a zone belongs to the library. Happy to open that on `jolt-lang/time` if useful.

One symptom there is worth flagging on its own: a consuming project's namespace that touches `java.time` is **dropped from its test suite rather than failing it**, and the runner exits 0. A downstream suite went from 29 tests / 78 assertions to 26 / 67, green. The only tell was the count.

burinc:
**Correction to my comment above.** I wrote that `jolt-lang/time` has not declared its `:jolt/provides`. That is wrong — it declared them on 2026-09-01 in `33c5e983` ("Declare the java.time classes jolt.time provides", #11), covering `Locale`, `DateTimeFormatter`, `ZoneId` and the rest. I read a local checkout that was four weeks stale instead of the remote. Apologies.

The library failures are real, and the cause is one step over.

**`jolt-lang/db` and `jolt-lang/transit-jolt` both pin `time` at `70dfb798`** (2026-07-29), which predates the declaration:

```
pinned:    70dfb798  2026-07-29  Declare the two forward-referenced vars (#3)
           -> no :jolt/provides at that tree, verified
declares:  33c5e983  2026-09-01  Declare the java.time classes jolt.time provides (#11)
```

So they resolve a `time` from before the migration and get no provider for `Locale` or `DateTimeFormatter`. `library time` passed in the same run, because the gate checks out that repo's default branch, which has the declaration.

`jolt-lang/db`'s deps.edn even carries the comment "jolt-lang/time is needed even though nothing here asks for java.time" — the dependency is there precisely to supply the classes, and the pin is what defeats it.

The fix is a pin bump in each consumer rather than anything in `time`. `transit-jolt` has not been touched since 2026-07-29 and also declares no `:jolt/provides` of its own.

**What this does not change, and arguably sharpens.**

The library gates still caught a real break that a push to `main` cannot surface, because those gates live only in `release.yml` behind `push: tags: v*` and `workflow_dispatch`, while `tests.yml` has none of them. A stale transitive pin is exactly the kind of breakage nobody discovers until a release runs — and it is arguably a better argument for a nightly than a missing declaration would have been, since a declaration gets fixed once and a stale pin can reappear in any consumer at any time.

It also makes the case for running the library matrix daily rather than only at tag time, independently of whether the nightly artifact itself is wanted.

**And a correction to my own claim about what I verified.** My local reproduction of `No dependency provides java.time.ZoneId` was run against a stale checkout of `time` as well, so it demonstrated the same pin-shaped problem rather than a missing upstream declaration. The CI failures stand on their own; that local repro does not add to them.

yogthos:
Merged with a fix commit on top (46eadbb4); its message has the details. Short version of what the review found: softprops/action-gh-release v2 looks the release up by tag, so from the second night on the matrix rows would have edited the live published nightly in place before the gates ran; the build rows' shallow checkout has no tags, so the nightly's binary would have reported a bare short sha, which :jolt/min-version reads as a version when it starts with a digit; and a vnightly tag at main's head is the nearest tag from main, so every clone that fetched it would have reported jolt --version as vnightly. The nightly is now published only by publish-nightly after the gates (stage draft, delete old, publish, move tag), a meta job decides the run kind and version once, and tools/version.sh is the single git describe. Thanks for the PR and the write-up, both made the review straightforward.

Issue #796, opened 2026-08-31
A portable-bytecode build reports itself as Linux: sa-os-family reads a machine tag that carries no OS


burinc:
`sa-os-family` derives the OS by substring-matching the Chez machine tag:

```scheme
;; host/chez/scheme-adapter-runtime.ss
(define (sa-os-family)
  (let ((m (sa-host-tag)))
    (cond ((or (sa-tag-contains? m "osx") (sa-tag-contains? m "macos")) 'macos)
          ((or (sa-tag-contains? m "nt") (sa-tag-contains? m "windows")) 'windows)
          (else 'linux))))
```

That works for a native tag. `tarm64osx` contains `osx`, `ta6nt` contains `nt`.

It cannot work for a portable-bytecode tag. `pb`, `pb64l` and `tpb64l` name the
threading, the word size and the endianness, and deliberately name no OS,
because the same bytecode is meant to run on any of them. The tag has no answer
to give, so the `else` branch fires and **every bytecode build calls itself
Linux — including one running on macOS.**

## Repro

No cross-compiler needed. This translates the `cond` verbatim and runs it
against real machine types:

```clojure
(ns os-family (:require [clojure.string :as str]))

(defn os-family [tag]
  (cond (or (str/includes? tag "osx") (str/includes? tag "macos")) :macos
        (or (str/includes? tag "nt") (str/includes? tag "windows")) :windows
        :else :linux))

(defn -main [& _]
  (doseq [tag ["tarm64osx" "ta6nt" "ta6le" "pb" "pb64l" "tpb64l"]]
    (println (format "  %-10s -> %s" tag (name (os-family tag))))))
```

```
  tarm64osx  -> macos
  ta6nt      -> windows
  ta6le      -> linux
  pb         -> linux     <-- on a Mac
  pb64l      -> linux     <-- on a Mac
  tpb64l     -> linux     <-- on a Mac
```

On a real `tpb64l` build running on a Darwin host,
`(System/getProperty "os.name")` returns `"Linux"` and `os.arch` returns
`"tpb64l"`.

## Why this is more than `os.name`

`sa-os-family` is the single place the rest of the host asks about the OS, and
its own doc comment lists the callers. Each takes a different value on Linux:

| caller | Darwin | Linux |
|---|---|---|
| `proc-SIGCHLD` (process.ss) | 20 | 17 |
| `proc-EAGAIN` (process.ss) | 35 | 11 |
[...]

PR #829, opened 2026-09-03
fix(adapter): a native iOS build is not Linux either

burinc:
## The gap #796 left

#796 fixed the portable-bytecode half of `sa-os-family`: `pb` tags carry no OS,
so the else-branch called every bytecode build Linux, and the fix probes the
filesystem when the tag has nothing to say.

The native half still falls through. Chez has four iOS machine tags, `a6ios`,
`arm64ios`, `ta6ios` and `tarm64ios`, and `BUILDING` documents `tarm64ios` as
the iOS cross-target. Not one contains `osx`, `macos`, `nt` or `pb`:

```scheme
(define (sa-os-family-for-tag m)
  (cond ((or (sa-tag-contains? m "osx") (sa-tag-contains? m "macos")) 'macos)
        ((or (sa-tag-contains? m "nt")  (sa-tag-contains? m "windows")) 'windows)
        ((sa-tag-contains? m "pb") (sa-probed-os-family))
        (else 'linux)))          ; <- every iOS tag lands here
```

So an iOS build answers `'linux` on a Darwin system.

## Why it matters beyond `os.name`

`sa-os-family` is the single place host logic asks about the OS, so this is
`SIGCHLD` 17 instead of 20, `EAGAIN` 11 instead of 35, `O_NONBLOCK`, `LC_TIME`,
the `struct stat` offsets, the chmod and entropy fallbacks, and the link
libraries in `build.ss`, all at once.

The one that bites first is #796's: `jolt.nrepl` computes `sock-cloexec` as
`0x80000` unless the OS is macOS or Windows, so an iOS build hands Darwin's
`socket()` the Linux-only `SOCK_CLOEXEC` and cannot bind at all.

## The change

`"ios"` joins the macos branch rather than getting a branch of its own. iOS is
Darwin, so every constant above is already the right one, and the contract is
three symbols with nowhere to put a fourth.

The substring is exact rather than lucky. Those four are the only tags in the
Chez tree containing `ios`, checked across `s/*.ss`, `c/*.c`, `configure` and
`BUILDING`, and a tag that contained it by accident would have to be an `osx`
tag, which wants `'macos` anyway.

## Tests

Through `sa-os-family-for-tag`, which #800 split out for exactly this: the rows
can be pinned from a host that is not iOS and cannot be.

```
$ chez --script test/chez/host-derived-props-test.ss     # without the fix
FAIL: tarm64ios -> os macos
FAIL: a6ios -> os macos
host-derived-props: 65 checks, 2 failures

$ chez --script test/chez/host-derived-props-test.ss     # with it
host-derived-props: 65 checks, 0 failures
```

The arch and endian rows for both tags pass either way, so the OS derivation is
the only thing wrong.

`make test` passes on the final tree: `OK: ci gate passed (97 targets)`,
`OK: test gate passed (3 targets)`.

## Verification

Branched from and re-verified against `upstream/main` at `ed9bb1d3`
(2026-09-03), which was still the tip when this was opened. The `ios` tag
enumeration was taken from a Chez v10.4.1 checkout.

Not claimed: I have not run a native `tarm64ios` build end to end, so the
socket consequence above is #796's measurement carried forward through the same
code path, not a fresh reproduction on an iOS target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A native iOS build no longer reports itself as Linux.** #796 fixed the
+  portable-bytecode half of `sa-os-family` — a `pb` tag names no OS, so the
+  `else` branch called every bytecode build Linux, and it probes the filesystem
+  now. The native half still fell through: Chez's four iOS tags (`a6ios`,
+  `arm64ios`, `ta6ios` and `tarm64ios`, the last documented in `BUILDING` as
+  the iOS cross-target) contain none of `osx`, `macos`, `nt` or `pb`, so all
+  four reached that same `else` and called a Darwin system Linux. iOS is Darwin
+  and `ios` joins the macOS branch, which is the whole fix: that one function is
+  where the host asks what OS this is, so the wrong answer was `SIGCHLD` 17
+  instead of 20, `EAGAIN` 11 instead of 35, `O_NONBLOCK`, `LC_TIME`, the
+  `struct stat` offsets, the chmod and entropy fallbacks and the link libraries,
+  all at once — and, as in #796, `jolt.nrepl` handing Darwin's `socket()` the
+  Linux `SOCK_CLOEXEC`, which cannot bind. `jolt build --target tarm64ios
+  --library` also linked the output with ELF's `-shared` rather than
+  `-dynamiclib -install_name`, because the build's target-side Darwin predicate
+  matched only `osx`; it takes both spellings now. Android needs no such branch
+  and gets none: it has no Chez tag of its own, cross-builds as `tarm64le`, and
+  Bionic is Linux for every constant these select — its divergence is the link
+  libraries, which no tag can express (`tarm64le` is glibc arm64 Linux too) and
+  the target pack owns. The gate pins the iOS and `tarm64le` rows through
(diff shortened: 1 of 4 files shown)


yogthos:
Merging. Thanks — the derivation is right, and Chez's own `examples/socket.ss` corroborates it: it puts `a6ios ta6ios arm64ios tarm64ios` in the same case arm as the `osx` tags, so upstream already treats them as one family.

Pushed a follow-up commit with three things the change was missing:

1. **`build.ss` still had the same bug.** `bld-tgt-osx?` is the build's own *target-side* Darwin predicate and matched only `"osx"`, so `jolt build --target tarm64ios --library` linked the output with ELF's `-shared` instead of `-dynamiclib -install_name` — an unloadable Darwin dylib. Same tag vocabulary, same fix.
2. **The CHANGELOG entry.** #796 has one and this extends it.
3. **Android, which needs no branch and now says so.** It has no Chez tag of its own — it cross-builds as `tarm64le` — and Bionic is Linux for every constant `sa-os-family` picks, so `'linux` was already right. Recorded as a decision rather than the else-branch's luck, with the `tarm64le` row pinned in the gate. The one place Android does diverge is the link libraries (no `-lrt`/`-luuid`/`-ltinfo`), which no tag can express since `tarm64le` is glibc arm64 Linux too, so that stays target-pack-owned as `tools/cross-compile/README.md` already documents.

Gate: 68 checks, 0 failures (was 65).

PR #834, opened 2026-09-03
Deftype class names answer in the JVM spelling at every name seam

yogthos:
Fixes the jolt-side bugs the typedclojure port surfaced. Beads jolt-anz9, jolt-gkxt, jolt-y82t, jolt-zhok, jolt-6fx6, jolt-arww.

A deftype in ns `rf.def-two` is `rf.def_two.R3` on the JVM. jolt registered it as `rf.def-two.R3` and only `Class/forName` and `instance?` knew both spellings, so `(resolve 'rf.def_two.R3)` was nil, `(extend-protocol P rf.def_two.R3 ...)` filed the impl under a tag no value carries (typedclojure's subtype table hit this as "No method subtypeA*-for-s" at the first dispatch), an `:import` of the munged form bound a second class token that was not `=` to `(class r)`, and the JVM's own record literal `#rf.def_two.R3{:a 1}` could not be read.

There is one canonicalizer now, `jch-registered-name` in class-hierarchy.ss (JVM spelling -> registered name, indexed off the class graph). The token interner asks it on first sight, so `resolve`, a class symbol in code, `:import` and `Class/forName` all intern the one token the type's values report; `register-method` and the record-literal reader map through the same function. `.getName`, `str`, `pr` and the record literal present the munged spelling, so a record round-trips through the JVM's reader. The written spelling still reads and resolves; that superset is a `:documented` entry in known-divergences.

Also from the same port:

- `extend-type`/`extend-protocol` on a bare name that resolves to nothing throws "Unable to resolve classname" like the JVM instead of registering a dead entry. The extending ns's class mapping (an `:import`) is consulted before the last-deftype-of-that-name index, as `Compiler.maybeClass` does. A dotted unmodeled name stays lenient (the `__register-class!` seam), documented.
- `import` takes a bare fully-qualified symbol spec (`(:import java.util.Date)`), which was silently dropped.
- `clojure.lang.Compiler/eval` and the `clojure.lang.LazySeq` constructor exist.
- A string method reached with too few args raises the reflector's IllegalArgumentException instead of an index fault from `list-ref`. The raw-condition half of jolt-arww was already fixed on main by #831's follow-up.

Gates: corpus 4994/5004 (0 new, 10 known), certify 0 new / 0 stale with both documented entries machine-checked on the JVM, `make test` green, `make libconformance` 47 ok / 0 better / 0 regressed (the 3 skips are the standing ones: process, core.typed, ansatz). The print-scaling test pinned the old hyphen spelling and now pins the JVM one.

Verified on the consumer: `typedclojure-jolt/bin/cf-gate.sh` passes on this branch with `subtype.jolt`, `env.jolt`, `reflect.jolt`, `requiring_resolve.jolt` and `safe_locals_clearing.jolt` removed, and `jolt_ext.clj` is redundant. The remaining two masks (`analyzer.jolt`, `contract_utils_platform_specific.jolt`) are host-neutral typedclojure fixes.


yogthos:
Local gates finished: `make test` green (all three targets), `make libconformance` 50 libraries: 47 ok, 0 better, 0 regressed, 3 skipped (the standing process/core.typed/ansatz skips). CI is green on both runs.

PR #852, opened 2026-09-04
Keep recur out of locking and with-out-str, and stop a bare try eating frames

yogthos:
Two bugs with one shared cause, both surfaced while testing #851.

## `locking` and `with-out-str` swallowed `recur` (jolt-h8wk)

The reference's `locking` is a `try`/`finally`, and its `with-out-str` goes through `binding`, which is one too — so a `recur` may not cross either. jolt runs both bodies in a thunk instead, because an OS mutex has thread granularity and a fiber is not a thread (`java/concurrency.ss:862-880` documents why a wind and the bare `monitor-enter`/`monitor-exit` halves were both rejected). That thunk quietly became the `recur` **target**:

| | reference | jolt before | jolt after |
|---|---|---|---|
| `(loop [] (locking o (recur)))` | compile error | **spins forever** | `Cannot recur across try` |
| `(loop [] (with-out-str (recur)))` | compile error | **spins forever** | `Cannot recur across try` |
| `(loop [i 0] (locking o (if (< i 3) (recur (inc i)) i)))` | `Cannot recur across try` | `Mismatched argument count` | `Cannot recur across try` |

Both macros now wrap their body in a bare `try` — which is what the reference's shape amounts to, and what `analyze-try` already reads. The thunk stays exactly as it was, so the fiber scheduling this exists for is untouched.

`lazy-seq` and `dosync` are **not** affected and are left alone: they wrap a thunk on the reference too, and a `recur` inside targets the thunk there as well (verified — `(loop [] (dosync (recur)))` hangs on the JVM).

## A bare `try` ate stack frames (jolt-je45)

The fix above is free only because `emit-try` emits a catch-less, finally-less try as its body and nothing else. But `:try` was not in `tail-transparent-ops`, so `emit` cleared `*tail?*` for it anyway. The call inside stayed a real Scheme tail call and lost its site, so TCO erased the frame with no name — and the trace stopped there:

```
(defn plain   [x] (boom x))        trace: boom, plain, cmd-run
(defn wrapped [x] (try (boom x)))  trace: boom, cmd-run       <-- wrapped and everything outside it, gone
```

Pre-existing, and independent of `locking` — any hand-written `(try x)` lost frames. Tail-transparency now reads the same two keys `emit-try` branches on. A try that *has* a catch or a finally is genuinely not tail-transparent and is untouched.

No emission change for `locking`: the body was in tail position before and still is, so the site is stored either way.

## Coverage

7 new corpus rows (2 refusals, 5 positives: locking returns its body, stays reentrant, a loop inside it recurs to itself; with-out-str still captures, and a loop inside it recurs to itself), 1 golden case, and 3 new trace-smoke assertions pinning the frames a bare try must keep.

## Not fixed here

`recur` is still accepted in **non-tail** position generally — `(loop [i 0] (+ 1 (recur (inc i))))` loops forever and silently discards the enclosing `+ 1`, where the reference reports `Can only recur from tail position`. jolt has no tail-position check for `recur` at all. That is a bigger change (it touches every special-form arm, and a wrong rule *rejects valid code*), so it is filed as jolt-c627 with its own test matrix rather than bolted on here.

## Gates

Both files are seed sources: remint (converged in 2 passes, 0 skipped) + gambit seed. `make ci` green, corpus throws-verified 194 → 196, trace smoke 65 → 68 assertions, selfhost and gambitseedcheck green.


yogthos:

PR #853, opened 2026-09-05
Match JVM Unicode ignore-case comparison

casselc:
# Match JVM Unicode ignore-case comparison

## Summary

Make `String.equalsIgnoreCase` and `String.compareToIgnoreCase` use the same
character-wise upper-then-lower fold already implemented for
`String/CASE_INSENSITIVE_ORDER`.

Previously the instance equality method lowercased ASCII only, and the compare
method lowercased whole strings but returned only `-1`, `0`, or `1`. Java's
contract compares folded characters and returns the difference of the first
unequal pair, or the length difference.

```clojure
[(.equalsIgnoreCase "É" "é")
 (.equalsIgnoreCase "É" nil)
 (.compareToIgnoreCase "a" "c")
 (.compareToIgnoreCase "Z" "a")]
;; before: [false <host error on nil> -1 1]
;; after:  [true false -2 25]
```

## Compatibility oracle

JVM Clojure 1.12 returns `[true false -2 25]` for the expression above. The
existing `String/CASE_INSENSITIVE_ORDER` corpus row already checks the same
comparison algorithm, including `"É"` / `"é"` and magnitude-bearing results.
The new corpus row independently runs the instance methods against the JVM.

## Red to green evidence

Base: `jolt-lang/jolt@f827badf7e1f540249377c98481f3f8a4a5b8909`

The unchanged stock implementation produces `[false -1 1]` for the non-null
three-case probe, while JVM Clojure produces `[true -2 25]`.

After the fix:

```text
unit gate: 1657/1657 passed
  57/57 str

Corpus parity: 5133/5143 evaluated cases pass
  crash: 0   NEW divergence: 0   known: 10
```

The ten corpus differences are the existing allowlisted set; this change adds
no new divergence.

## Validation

- based on `jolt-lang/jolt@f827badf7e1f540249377c98481f3f8a4a5b8909`
- current branch tip: `4fc9e6e8`
- `make unit` through Chez Scheme 10.4.1: PASS, 1657/1657
- `make corpus` through Chez Scheme 10.4.1: PASS, zero new divergences

## Provenance

Tracked as
[chucklehead-dev/jolt-aspect-packs#75](https://github.com/chucklehead-dev/jolt-aspect-packs/issues/75).
That ledger is provenance only; the reproducer and fix require no aspect
machinery.

## Scope

The patch moves the already-present case-insensitive comparator into the shared
String method layer, points the instance methods and static comparator at it,
adds host-unit and JVM-oracle corpus rows, and records the fix in the changelog.
It does not change `toUpperCase` or `toLowerCase`, whose locale-sensitive surface
is separate.


yogthos:
Reviewed and merged-ready from my side — I pushed one follow-up commit (f039f7a) to this branch rather than blocking on it, since it's the same two contracts this PR is about.

Two neighbours were left behind by the fix:

**`compareTo` still answered a sign.** With `compareToIgnoreCase` now reporting the real character difference, its case-sensitive twin reporting `-1` was two contracts for one pair of methods — and `(.compareTo "a" "c")` disagreeing with `(.compareToIgnoreCase "a" "c")` on the same input is how a caller finds out. JVM: `-2` and `2`, not `-1` and `1`.

**`regionMatches` folded a third way.** It used Scheme's `string-ci=?`, which is full Unicode case folding, not Java's per-character upper-then-lower one. On the JVM `equalsIgnoreCase` *is* `regionMatches(true, 0, other, 0, length)`, so those two cannot disagree — but here they did:

```clj
(.equalsIgnoreCase "I" "ı")            ; => true  (both upper-case to I)
(.regionMatches   "I" true 0 "ı" 0 1)  ; => false (Unicode folds them apart)
```

One JVM operation, two answers. The fold is now one function (`jvm-char-ci-diff`) that all four go through — `equalsIgnoreCase` and `regionMatches` as equality, `compareToIgnoreCase` and `CASE_INSENSITIVE_ORDER` as a difference — so a new ignore-case method has one place to reach for instead of a choice between three foldings.

Both new corpus rows verified against Clojure 1.12 (`[-2 2 -2 0 -7]`, `[true true true false]`); `make unit` green, `str` 59/59.

One thing I deliberately left alone: `clojure.core/compare` on strings, keywords and symbols still goes through `jolt-strcmp`, which is sign-only by a documented `-1/0/1` invariant that `register-compare-arm!` handlers share. So `(compare "a" "c")` is `-1` here and `-2` on the JVM. That's a real divergence but a much wider change than this PR, and it isn't in `known-divergences.edn` either — worth its own issue.

PR #854, opened 2026-09-05
Honor the Java collection toArray contract

casselc:
# Honor the Java collection toArray contract

## Summary

Implement the Java `Collection.toArray` result and destination-array contracts
for Jolt's modeled `ArrayList`, `LinkedList`, and `ArrayDeque`, which share the
same method table.

Both call shapes previously ignored their overload semantics and returned a
Clojure persistent vector. The destination overload now reuses a sufficiently
large object array, writes the null logical terminator while preserving later
cells, and allocates a replacement when the supplied array is too short.

```clojure
(let [a (java.util.ArrayList. [1 2])
      dst (object-array [:old :old :keep :tail])
      out (.toArray a dst)]
  [(str (class (.toArray a)))
   (identical? dst out)
   (vec out)])
;; before: ["class clojure.lang.PersistentVector" false [1 2]]
;; after:  ["class [Ljava.lang.Object;" true [1 2 nil :tail]]
```

## Compatibility oracle

JVM Clojure 1.12 produces the `after` result above. The JVM-oracle corpus row
also covers a fresh no-argument result and the undersized destination case:
the latter returns a new `[1 2]` array rather than reusing a one-cell input.

Jolt models reference arrays with one `object` kind, so it cannot preserve a
more-specific reference component class when allocating the undersized
replacement. It does preserve every array distinction the runtime represents;
primitive arrays are rejected rather than silently filled as object arrays.

## Red to green evidence

Base: `jolt-lang/jolt@f827badf7e1f540249377c98481f3f8a4a5b8909`

The stock probe above returns a persistent vector and reports that the supplied
destination was not reused. After the fix:

```text
unit gate: 1660/1660 passed
  16/16 hostctor

Corpus parity: 5134/5144 evaluated cases pass
  crash: 0   NEW divergence: 0   known: 10
```

The ten corpus differences are the existing allowlisted set; this change adds
no new divergence.

## Validation

- based on `jolt-lang/jolt@f827badf7e1f540249377c98481f3f8a4a5b8909`
- current branch tip: `7014b077`
- `make unit` through Chez Scheme 10.4.1: PASS, 1660/1660
- `make corpus` through Chez Scheme 10.4.1: PASS, zero new divergences

## Provenance

Tracked as
[chucklehead-dev/jolt-aspect-packs#76](https://github.com/chucklehead-dev/jolt-aspect-packs/issues/76).
That ledger is provenance only; the reproducer and fix require no aspect
machinery.

## Scope

The patch changes the shared modeled collection implementation, adds host-unit
and JVM-oracle corpus rows for `ArrayList`, `LinkedList`, and `ArrayDeque`, and
records the fix in the changelog. It does not add a new array representation or
claim component-type fidelity Jolt does not model.


yogthos:
Reviewed — the contract work here is right (the null logical terminator, preserving later cells, reallocating when the destination is short, and the `Object[]` component type all match the JVM). I pushed one follow-up commit (f6b8837) to this branch.

**`HashSet` had no `toArray` at all.** `toArray` is a `Collection` method, not a `List` one, so the set answers it on the JVM; here it was:

```
No matching field found: toArray for class java.util.HashSet
```

while every other modeled collection in the file now answers both overloads.

So `al-to-array` is now `jcoll-to-array`, taking a collection's element list rather than reaching into `ArrayList`'s backing vector — the fill loop walked `al-ref`, which only a `List` has. The list is what the two have in common, so the set shares the contract rather than a copy of it, and whatever gets modeled next needs only its own list. The one cost is that the fits-in-place case now builds a list it previously didn't; for a shim this size that isn't worth a second implementation to avoid.

Verified against Clojure 1.12 — fresh array, component type, reused destination with terminator, and reallocation all agree:

```clj
[(vec (.toArray (java.util.HashSet. #{1 2})))
 (instance? (Class/forName "[Ljava.lang.Object;") (.toArray (java.util.HashSet. #{1})))
 (let [h (java.util.HashSet. #{1 2}) d (object-array [:o :o :keep])] [(identical? d (.toArray h d)) (vec (.toArray h d))])
 (let [h (java.util.HashSet. #{1 2}) d (object-array 1)]              [(identical? d (.toArray h d)) (vec (.toArray h d))])]
;; jolt and JVM: [[1 2] true [true [1 2 nil]] [false [1 2]]]
```

`make unit` green, `hostctor` 17/17.

Also checked and left alone: `(.toArray coll nil)` raises `NullPointerException` here, matching the JVM (I'd assumed `jolt-cast-throw` would give a CCE — it doesn't). `HashSet` is still missing `iterator`, `addAll` and `containsAll`, which is pre-existing and outside this PR.

PR #855, opened 2026-09-05
Refuse recur outside tail position

yogthos:
Stacked on #852, which is stacked on #851. Merge in that order.

`recur` rebinds and jumps, so anywhere but tail position the surrounding expression is silently discarded. jolt had no tail-position check at all:

```clojure
(loop [i 0] (+ 1 (recur (inc i))))     ; jolt: loops forever, (+ 1 ...) never applies
                                        ; ref:  Can only recur from tail position
```

That is a miscompile, not just a missing diagnostic — the enclosing computation is dropped rather than the program being refused.

## Why a pass, not threaded through the analyzer

The obvious implementation threads a `:tail?` flag through every `analyze-*` arm. That is the risky one: each arm needs its own rule, and **one wrong rule rejects valid code** — worse than the permissiveness it replaces.

Tail position is a property of the analyzed tree, so this checks the tree instead. The default is "children are not tail", so an op the pass does not name can only ever be restrictive about a `recur` nested inside it, never wrong about one in tail position — the uncertainty falls in the safe direction.

Only three ops pass tail position down — `if` to its branches, `do` to its last form, `let` to its body — and they are the same three the back end's `tail-transparent-ops` already names for the same question, so the rules are cross-checked against a working implementation rather than invented. `case`, `cond`, `when`, `or`, `and` and `letfn` are macros over exactly those and need no rule of their own.

`:fn` and `:loop` are their own recur targets, checked at their own analysis sites. A loop's binding *inits* still belong to the enclosing target, so `(loop [i 0] (loop [j (recur 1)] j))` is caught.

## Position

A `:recur` node now carries its source position. The pass runs after the body is analyzed, when the position box has moved on, so without it the report named the enclosing `loop`. jolt now reports `6:12` where the reference reports `pos.clj:6:12` for the same program.

## Coverage

18 corpus rows: 6 refusals and **12 positives**. The positives are the point — they are what would catch over-rejection: `recur` as the last form of a fn body, a loop body, both `if` arms, a `do`'s ret, a `let` body, through nested lets, in a multi-arity fn, and in `when` / `cond` / `case` / `letfn` bodies. Plus one golden case.

The remint is the broadest evidence: it recompiles the whole compiler and all of `clojure/core` under the new rule and converged in 2 passes with 0 forms skipped.

## Gates

`analyzer.clj` is a seed source, so this carries a remint + gambit seed. `make ci` green, corpus throws-verified 196 → 202, certify 0 new / 0 stale, irvalidate green (the `:recur` node's new `:pos` is an annotation key the schema already allows).

## A latent bug this exposed, fixed in the second commit

`make ci` passed all 98 targets on the first commit while it was **breaking eight malli namespaces** — only `make libconformance` caught it (`load-fail` 1 → 9, every one `Can only recur from tail position`).

The cause was a false positive, and a pre-existing jolt bug the new check made reachable. A conditions map wrapped the body in `(let [% (do body)] %)` whether or not there were any `:post` conditions, which puts the body in a binding *init* — not tail position. malli's `validate-times` is exactly that shape:

```clojure
(defn validate-times [n schema v]
  {:pre [(pos? n)]}
  (let [res (m/validate schema v)]
    (if (or (= 1 n) (not (true? res))) res (recur (dec n) schema v))))
```

The reference emits the `%` wrapper only for `:post`; jolt now matches.

Tail *calls* were never affected — measured, not assumed: 3e6 deep self-calls through the var run at baseline RSS with the wrapper and without it, because the back end collapses `(let [% X] %)` to `X`. Only `recur` saw it, since the analyzer checks the tree before that collapse.

`make libconformance` on the final tree: **47 ok, 0 better, 0 regressed, 3 skipped**.


yogthos:

PR #857, opened 2026-09-05
Give compile errors a kind, a position and a real throwable

yogthos:
R1 of the error-reporting epic. Stacked on #855 → #852 → #851; merge in that order.

A compile diagnostic was a bare Scheme string — `(throw (str "jolt/uncompilable: " why))`. Caught, that gave:

```
(class e)               => String            not a throwable
(ex-message e)          => nil               a catch cannot read it
(instance? Exception e) => false
```

The message survived only because the uncaught reporter `pr-str`'d the raw value, so the *report* looked fine while nothing could act on it. 29 raise sites now go through `analysis-error`, which throws an `ex-info` carrying a namespaced `:kind`, the innermost positioned form's `file:line:column`, and the thrower's own ex-data.

## The registry gate

`test/conformance/error-kinds.edn` documents all 18 kinds; `make errorkinds` holds it and the sources to each other **in both directions** — an unregistered kind is an error with no documented meaning, and a registered kind nothing raises is a doc rotting into decoration. Both directions are mutation-tested.

## Messages that named the wrong cause

| | before | after |
|---|---|---|
| `(def :foo 2)` | `def name with map metadata` | `First argument to def must be a Symbol` |
| `(let [a 1 b] a)` | `IndexOutOfBoundsException` | `Bad binding form, expected matched symbol expression pairs` |
| `(defn f 5 6)` | `Don't know how to create ISeq from: java.lang.Long` | `Parameter declaration 5 should be a vector` |
| `(if)` | `Wrong number of args (0) passed to: if` | `Too few arguments to if` |

**Two of those were not in the analyzer at all.** `destructure` and `defn`'s arglists walk both step through their forms in pairs and read past the end on a malformed one, so every `let`/`loop`/`for`/`doseq` and every `defn` hit the raw host fault before the analyzer ever saw the form. Converting the analyzer arm that *looked* responsible changed nothing; running the binary is what showed it.

## ex-data

`as-analysis-diagnostic` rebuilt the ex-info with only `:jolt/error`, discarding whatever the thrower attached — a macro raising `(ex-info "m" {:orig true})` reported no `:orig`, and the reporter then printed the compiler's own map in its place. The original ex-data is kept and `:jolt/error` sits beside it; the report shows the user's map and never the compiler's.

## Position for macro-raised diagnostics

A diagnostic raised from `clojure.core` carries a kind but no position, since it cannot reach the analyzer's position box. Those now have the position filled in rather than being passed through — *already a diagnostic* means don't **overwrite** what it has, not don't **fill in** what it lacks. Without that, `(defn f [] (let [a 1 b] a))` named the `defn`'s line 3 where the reference names the `let` on line 4. Caught by reading the golden diff.

## Gates

`make ci` green (99 targets, `errorkinds` new), corpus 5152/5162 with 0 new divergences, certify 0 new / 0 stale, **`make libconformance` 47 ok / 0 regressed** — which matters here because the `destructure` guard fires on every binding vector in every library. `make unit` gained two assertions pinning the new contract (class is `ExceptionInfo`, `:kind` is carried) rather than merely dropping the two that pinned the old bare string.

18 of the 33 golden reports changed, every one an improvement; the diff is in the commit.


yogthos:

PR #861, opened 2026-09-06
Reduce confined FFI arena bookkeeping

casselc:
# Reduce confined FFI arena bookkeeping

## Summary

Confined arenas currently use the same compare-and-swap bookkeeping as shared
arenas even though a confined arena has exactly one legal mutator: its owner
thread. For short lexical scopes, the bookkeeping can cost more than the native
allocation and free it protects.

This change keeps every arena state in an atom so `arena-open?` retains the
same cross-thread visibility, but specializes the confined owner path:

- attach and close publish with `reset!` instead of CAS-loop
  `swap-vals!`/`reset-vals!`;
- an empty arena skips cleanup traversal and failure bookkeeping;
- a group containing one ordinary allocation frees its raw pointer and forgets
  the usable address directly; and
- the general cleanup path passes values to one helper instead of allocating a
  closure for every release.

Shared, automatic, and global arenas retain their atomic update/exchange paths.
This does not pool allocations or change arena lifetime, ownership, or release
order.

## Correctness envelope

The exclusive update is sound because all confined-arena mutation is guarded by
the owner-thread check. A foreign thread may observe `arena-open?`, but cannot
allocate in or close the arena. State remains an atom for that observer.

Native allocation may re-enter owner-thread code and close the arena between
the initial open check and attachment. The confined attach path therefore
re-reads the state immediately before publishing; when it finds a reentrant
close, the existing `on-lost` rollback frees the new allocation and the call
raises.

Cleanup semantics are retained:

- views run before callables, followed by blocks in reverse allocation order;
- the general path continues after a cleanup failure and rethrows the first
  failure after size metadata is removed; and
- the one-block path frees the raw allocation while forgetting the possibly
  aligned usable address in `finally`.

New deterministic tests cover a rejected non-owner close while the arena stays
open, reentrant owner close during allocation with exactly-once rollback,
single-block free failure with size cleanup, and reverse multi-block release.
The existing suite continues to cover shared close/attach races, wrong-thread
allocation, idempotent close, views, callables, automatic arenas, and cleanup
continuation.

I also modeled one confined-arena transition in Chiasmus over `Open`/`Closed`,
owner/other caller, use/attach/close, and zero through four existing blocks.
The reference-versus-candidate observable-divergence query was UNSAT. A mutant
that allowed another thread to close was SAT with the expected
`Open / Other / Close` witness, and an owner fifth attachment was SAT with five
owned blocks and none released. This is a bounded state-machine check, not a
proof of native free behavior or release order; the runtime tests are the oracle
for those properties.

## Precommitted performance target

Before implementation, the target and measurement method were recorded in
[chucklehead-dev/jolt-aspect-packs#100](https://github.com/chucklehead-dev/jolt-aspect-packs/issues/100):

- improve empty and one-block confined-arena medians by at least 25%;
- keep raw allocation and amortized 100-block medians within 10% of base; and
- keep candidate empty/raw at or below 2.30x and one-block/raw at or below
  6.50x.

The diagnostic runs 30,000 eight-byte allocations, performs a checked native
write/read on every block, warms the raw, empty, one-block, and multi-block
paths, retains all five monotonic-clock samples, and reverses case order on
alternating rounds. Base/candidate/base runs were pinned to the same CPU. No
best-of selection or retry was used.

Exact current base: `2852ad3e7dc9ca97ee03f5c57b9f219883d97c47`

Exact candidate: `9cadf24cb3dcec14cfa98d4a6857d16e3279b1c8`

Medians in milliseconds:

| mode / case | base A | candidate B | base A2 | candidate vs base range | gate |
|---|---:|---:|---:|---:|---:|
| source raw control | 25.82 | 27.78 | 25.72 | +7.6% to +8.0% | pass |
| source empty arena | 68.52 | 32.17 | 68.74 | -53.0% to -53.2% | pass |
| source one block | 190.48 | 120.31 | 192.83 | -36.8% to -37.6% | pass |
| source 100 blocks | 86.47 | 86.21 | 85.48 | -0.3% to +0.9% | pass |
| AOT raw control | 28.12 | 25.99 | 23.92 | -7.6% to +8.7% | pass |
| AOT empty arena | 84.26 | 25.24 | 69.29 | -70.0% to -63.6% | pass |
| AOT one block | 225.75 | 104.56 | 168.98 | -53.7% to -38.1% | pass |
| AOT 100 blocks | 77.65 | 73.87 | 73.99 | -4.9% to -0.2% | pass |

Candidate ratios are 1.16x empty/raw and 4.33x one-block/raw in source mode,
and 0.97x and 4.02x AOT.

The earlier pre-v0.8.2 AOT baseline had a 22.15 ms raw median and therefore a
24.36 ms absolute control threshold. The current candidate's 25.99 ms misses
that historical absolute threshold, as does one of the exact current base
brackets; the symmetric current-base regression gate passes. The candidate
still passes the historical absolute targets for empty, one-block, and
100-block arenas. This is why the PR treats the exact-current paired controls,
not an old compiler's absolute raw timing, as the regression decision.

<details>
<summary>Retained target-case samples (nanoseconds)</summary>

```text
source base A
raw    [25901974 25451071 25374111 26887097 25820613]
empty  [70966341 68792872 68285111 67661888 68519327]
one    [190482330 186843827 190414137 196142720 195832684]
100    [96894320 86924184 86274067 86474006 85465138]

source candidate B
raw    [27764337 27780737 27580924 30266774 29839849]
empty  [34837050 31920242 31100985 32172655 32498599]
one    [138124963 120309929 121094656 119511096 118063067]
100    [86211993 85346632 86168275 107018476 88849900]

source base A2
raw    [25590418 26167026 25391724 26011079 25715757]
empty  [73655676 67242576 67900554 68740119 69149238]
one    [191961427 188578155 196388969 205210042 192825389]
100    [84063553 83816583 86597224 86169606 85478937]

AOT base A
raw    [62187259 28689354 24158674 28120266 23361722]
empty  [84710111 84937744 72940827 84257308 71279036]
one    [243496818 269084053 184705028 225748623 200334894]
100    [77648911 73020726 81035417 72941234 80964314]

AOT candidate B
raw    [29039480 25994711 22968076 26008732 23193540]
empty  [27456220 26001123 24799749 25237908 24988259]
one    [104561948 99171154 111028775 101698158 105289537]
100    [70863617 74469426 70484856 78775867 73874474]

AOT base A2
raw    [25764606 23922372 22326858 26976051 22398377]
empty  [70546902 69285097 67978757 78503619 68809529]
one    [173523342 167843536 168065730 168983574 172929093]
100    [73988419 76918077 71176601 75081171 73675073]
```

</details>

## Validation

- current `origin/main`: `2852ad3e7dc9ca97ee03f5c57b9f219883d97c47`
- branch tip: `9cadf24cb3dcec14cfa98d4a6857d16e3279b1c8`
- `make -j1 ffi` through Chez Scheme 10.4.1 — PASS:
  `47/47`, `63/63`, `46/46`, layout OK, `18/18`, aggregate OK,
  scoped OK, arena OK, native-error `41/41`
- exact-tip upstream workflows — PASS: [aggregate test and Windows dependency
  paths](https://github.com/jolt-lang/jolt/actions/runs/34001751867),
  [Linux and macOS](https://github.com/jolt-lang/jolt/actions/runs/34001751898)
- independent Claude review — approve with edits; its requested distinct-path
  warmups and non-owner-close regression are incorporated in this tip

## Provenance

The motivating measurements, precommitted target, and full investigation log
are tracked in
[chucklehead-dev/jolt-aspect-packs#100](https://github.com/chucklehead-dev/jolt-aspect-packs/issues/100).
That issue is provenance only; this patch, benchmark, and tests run on stock
Jolt without the aspect/compiler tooling.

## Scope

The branch changes `stdlib/jolt/ffi.clj`, adds one standalone Jolt-only
diagnostic under `bench/`, extends the existing arena test and benchmark docs,
and adds an Unreleased changelog entry. It changes no public API, native
allocator, lifetime contract, or shared-arena synchronization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Confined FFI arenas avoid shared-arena compare-and-swap bookkeeping on their
+  owner-only attach and close paths. Empty and single-allocation lexical arenas
+  also skip cleanup work they do not need, without changing arena lifetime or
+  release-order semantics.
+
 ## [0.8.2] - 2026-09-05
 
 Errors say what went wrong, where, and can be caught. A compile error is a framed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Confined FFI arenas avoid shared-arena compare-and-swap bookkeeping on their
-  owner-only attach and close paths. Empty and single-allocation lexical arenas
-  also skip cleanup work they do not need, without changing arena lifetime or
-  release-order semantics.
+- Confined FFI arenas avoid the shared-arena compare-and-swap LOOP on their
+  owner-only attach and close paths, publishing with one compare-and-set!
+  instead. Empty and single-allocation lexical arenas also skip cleanup work
+  they do not need, without changing arena lifetime or release-order semantics.
+
+### Fixed
+
+- A confined FFI arena shared between fibers on one carrier no longer loses
+  allocations. Fibers share their carrier's thread id, so they all pass the
+  arena's owner check; the attach and close paths publish atomically rather
+  than assuming one mutator.
 
 ## [0.8.2] - 2026-09-05
 
--- a/bench/README.md
+++ b/bench/README.md
@@ -27,6 +27,13 @@ and the [Computer Language Benchmarks Game](https://benchmarksgame-team.pages.de
 The benchmarks are portable Clojure, so they also run on JVM Clojure for an
 absolute reference.
 
+`ffi_arenas.clj` is a separate Jolt-only diagnostic because the JVM has no
+`jolt.ffi` reference implementation. Run it from this directory with
+`../bin/jolt -Srepro -Sdeps '{:paths ["."]}' -m ffi-arenas`. It retains all five
+monotonic-clock samples and alternates scenario order; compare exact base and
+candidate runs on the same host rather than treating one absolute timing as a
+portable threshold.
+
 ## Benchmarks
 
 | Benchmark | Axis | Pass it exercises | Source |
--- a/bench/ffi_arenas.clj
+++ b/bench/ffi_arenas.clj
@@ -1,0 +1,57 @@
+;; Confined-arena bookkeeping benchmark. Jolt-only: there is no JVM FFI
+;; reference for this API. Retain every sample and alternate scenario order so
+;; drift is visible; compare exact base/candidate runs on the same host.
+(ns ffi-arenas
+  (:require [jolt.ffi :as ffi]))
+
+(defn- touch [p]
+  (ffi/write p :int64 42 0)
+  (when-not (= 42 (ffi/read p :int64))
+    (throw (ex-info "FFI round-trip failed" {}))))
+
+(defn- raw [n]
+  (dotimes [_ n]
+    (let [p (ffi/alloc 8)]
+      (try (touch p) (finally (ffi/free p))))))
+
+(defn- empty-arenas [n]
+  (dotimes [_ n]
+    (let [a (ffi/confined-arena)]
+      (ffi/close-arena a))))
+
+(defn- grouped [n batch]
+  (dotimes [_ (quot n batch)]
+    (let [a (ffi/confined-arena)]
+      (try
+        (dotimes [_ batch] (touch (ffi/alloc a 8)))
+        (finally (ffi/close-arena a))))))
+
+(defn- time-call [f]
+  (let [start (System/nanoTime)]
+    (f)
+    (- (System/nanoTime) start)))
+
+(defn -main [& args]
+  (let [n (if (seq args) (Integer/parseInt (first args)) 30000)
+        cases [[:raw #(raw n)]
+               [:empty-arena #(empty-arenas n)]
+               [:arena-per-allocation #(grouped n 1)]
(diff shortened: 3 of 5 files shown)


yogthos:
Reviewed and merging — with one follow-up commit pushed to the branch, because the correctness envelope's premise does not hold under fibers.

`arena-usable!` compares `jolt.host/thread-id`, which is `get-thread-id` — the *carrier's* id. Every fiber on one carrier reports the same id, so they all pass a confined arena's owner check, and preemption is polled at procedure entries, so a switch lands between the read and the publish as readily as anywhere else. One owner thread is not one mutator.

200 fibers allocating into one confined arena on one carrier, with the branch as it stood: 29 of 10000 blocks end up recorded. The rest are allocated, attached to a group that no longer exists, and never freed. The close side drops a sibling's block the same way, and an attach that lands after a close re-opens the arena around a pointer that has already been freed.

The fix keeps everything the branch is actually here for. The CAS *loop* is what the fast path exists to skip, not the CAS: `compare-and-set!` is the same single mutex acquisition `reset!` was and allocates no `[old new]` pair, so only the rare loss falls back to `swap-vals!`/`reset-vals!`. The closure elision, the empty-group skip and the one-block path are untouched.

Two rows added, each failing without its own half:
- sibling fibers on one carrier lose no allocation (the attach side, real fibers)
- a sibling allocation between a close's read and its publish is not dropped — the close-side window is a couple of instructions wide, so the row drives the interleaving from a validator on the state atom, which runs after the read and before the publish

Re-measured on your own `bench/ffi_arenas.clj`, medians of interleaved builds, 30000 8-byte allocations, branch tip vs. tip+fix: empty arena 22.24 -> 23.12ms, arena-per-allocation 82.78 -> 80.71, per-10 61.28 -> 60.72, per-100 54.33 -> 54.19, per-1000 54.55 -> 54.16. The raw control, which touches no arena code at all, moved 18.89 -> 19.43 over the same builds, and a variant with the close left as `reset!` measured empty arena at both 22.23 and 23.15 — so nothing here resolves as a cost of the change.

`make ffi` green.

PR #876, opened 2026-09-07
Complete jolt's commands and a project's tasks, in zsh, bash and fish

### LINKED ISSUE #878: No shell completion for jolt, and the fast way to add it is not babashka's way
`jolt <TAB>` offers nothing, in any shell. Greps for `compgen`, `COMP_WORDS`,
`bash-completion`, `--completions` and `_jolt` across the tree return nothing,
`find` turns up no completion file, and neither README nor CONTRIBUTING mentions
it. `bin/` holds the launcher and a shim. So this is a gap rather than something
to extend.

It matters a little more for jolt than for most CLIs, because `jolt tasks` reads
`bb.edn` as well as `deps.edn`. A project can move off babashka's runner by
changing the command people type — and the thing they immediately lose is
`bb <TAB>`, which the babashka book documents and a lot of projects lean on.

### The part I think is worth deciding before code

The obvious move is to copy babashka, which since 1.13.219 has
`bb org.babashka.cli/completions snippet --shell zsh` emit a zsh function that
calls `bb` back on **every TAB press** to compute the candidates for the line
so far.

That design does not port, and the reason is startup. Measured on an M-series
mac, release binary, five runs each:

| command | real |
|---|---|
| `jolt version` (resolves no project at all) | 0.21-0.22s |
| `jolt -e '(println 1)'` | 0.22-0.24s |
| `jolt tasks` | 0.21-0.27s |

The floor is the runtime coming up rather than a project resolving, which is why
`version` costs what `tasks` costs. A per-press callback therefore puts about a
quarter of a second between TAB and the first candidate. For comparison,
`bb tasks` measures 97ms here and babashka's own completion callback 100-160ms —
and babashka is the tool whose design this would be copying.

(I see #874 is measuring and improving cold start, so that table is a number
you are actively moving. It doesn't change the shape of the argument below, but
it is fair to read those figures as "at 92fef2ce".)

### What I'd suggest instead

Split the two halves by how often they change:

- **jolt's own commands and options change when the binary does**, so bake them
  into the snippet at the moment it is generated. A completing shell then pays
  nothing for them, ever.
- **A project's tasks change when its `deps.edn` or `bb.edn` changes**, so have
  the snippet cache them against those two mtimes and call back only when one
  moves.

Under zsh that path can use `zsh/stat` and `$(<file)`, so it forks no process at
all: **0.4ms** on a warm press. The callback then runs about once per edit of a
config file rather than once per keystroke.

The cost is that per-task **argument** completion is out of reach, since that
genuinely needs the line. jolt has no per-task option metadata today (babashka
gets it from `:exec-fn`/`:cmd` routing through `babashka.cli`), so there is
nothing to offer yet, and the design doesn't foreclose a line-aware callback
later for tasks that grow one.

### Open question

Whether `completions` should be a command at all, or whether the snippets belong
in `share/` as files the installer places. I leaned to the command because it
needs no install-path handling and behaves the same from Homebrew, the curl
installer, or a checkout — but I'd rather have your call than my guess.

### Status

PR #876 implements the above for zsh, bash and fish, with a gate
(`make completionssmoke`, 24 checks) and hand-verification in two unrelated
projects. It stacks on #875/#877, because a multi-line `:doc` otherwise puts a
phantom name in the candidate list.

Happy to rework the shape, or to drop it if you'd rather completion lived
outside the binary.

### Verified against

`main` at 92fef2ce (tag v0.8.4), 2026-09-07, macOS arm64. babashka figures from
1.12.196 and 1.13.220.

burinc:
Closes #878.

`jolt <TAB>` offers nothing today, in any shell — greps for `compgen`,
`COMP_WORDS`, `--completions` and `_jolt` across the tree come back empty. This
adds `jolt completions zsh` (or `bash`, or `fish`), which prints a function to
source. Under zsh each task carries its `:doc` as the description.

> **Stacked on #875.** This branch contains that commit, because the completion
> depends on its `:doc` truncation: without it a multi-line docstring puts a
> phantom name in the candidate list. Merge #875 first and this shrinks to its
> own commit. Happy to rebase if you'd rather it stood alone.

### The design is not babashka's, and that is the part worth your eye

The obvious move is to copy babashka, which since 1.13.219 emits a zsh function
that calls `bb` back on **every TAB press** to compute candidates for the line
so far.

That doesn't port, because of startup. Measured here, release binary, five runs:

| command | real |
|---|---|
| `jolt version` (resolves no project at all) | 0.21-0.22s |
| `jolt -e '(println 1)'` | 0.22-0.24s |
| `jolt tasks` | 0.21-0.27s |

The floor is the runtime coming up, not a project resolving — which is why
`version` costs what `tasks` costs. A per-press callback therefore puts about a
quarter second between TAB and the first candidate. For scale, `bb tasks`
measures 97ms on this machine, and babashka is the tool whose design this would
be copying.

So the two halves are split by how often they change:

- **jolt's own commands and options change when the binary does**, so they are
  baked into the snippet as it is generated, and cost a completing shell nothing.
- **A project's tasks change when its `deps.edn` or `bb.edn` changes**, so the
  snippet caches them against those two mtimes and calls back only when one moves.

Under zsh that path uses `zsh/stat` and `$(<file)` and forks no process at all.
A warm press measures **0.4ms**. `JOLT_COMPLETION_NO_CACHE=1` bypasses it.

I saw #874 while checking this was still current, so the 0.22s figure is a
number you are actively working on. The split doesn't depend on it: 0.4ms beats
any process spawn by orders of magnitude, and a faster cold start makes the
cache-miss path cheaper rather than making the design wrong. It does mean the
paragraph above should be read as "as of 92fef2ce".

The cost of the split is that per-task **argument** completion is out of reach,
since that genuinely needs the line. jolt has no per-task option metadata today
(babashka gets it from `:exec-fn`/`:cmd` through `babashka.cli`), so there is
nothing to offer yet, and nothing here forecloses a line-aware callback later.

**One design question I'd want your call on:** whether `completions` should be a
command at all, or whether the snippets belong in `share/` as files the
installer places. I went with the command because it needs no install-path
handling and works the same from a Homebrew install, the curl installer, or a
checkout — but it is your call and I'll rework it.

### `jolt completions tasks`

The callback, and useful alone: one `name<TAB>doc` per listable task, the
machine-readable form of what `jolt tasks` writes for a person. Both go through
the new `jolt.tasks/listable`, so the listing and TAB cannot disagree about
which tasks exist.

They differ on one point by intent. A task sharing a built-in's name is offered
only when it wins that name with `:override-builtin`, and the snippet drops the
command it displaced. A description is a promise about what the word will do,
and for a task that loses to a command the answer is the command.

### Three bugs found by running a real zsh, not by reading the output

None is visible to `zsh -n`:

- A file saved as `_jolt` on `$fpath` is **autoloaded**, and zsh runs the FILE
  as the function body. A snippet that only defines `_jolt` completes nothing on
  the press that loaded it. It checks `zsh_eval_context` now and either calls the
  function or registers it with `compdef`, so sourcing and fpath both work.
- `_describe` splits `value:description` at the first colon and eats
  backslashes, so both are escaped in both halves. Task names carry colons
  routinely (`build:linux`, `release:macos`); unescaped, such a name is cut at
  the colon and vanishes behind the command it then collides with.
- `_describe` does **not** keep the first of a repeated value, so offering both
  a task and the command it overrides showed the command's description on the
  one name where the task is what runs. Dropping the displaced command
  explicitly is the fix; ordering is not.

That third one survived a green gate, which is why the gate grew a group that
sources the generated zsh function with `_describe` stubbed and reads the
candidates back. bash has no description column, so a candidate carrying the
wrong description was invisible to every check that existed at the time.

### Gates

```
make completionssmoke   24 passed, 0 failed
make taskssmoke         60 passed, 0 failed
make scriptsmoke        33 passed, 0 failed
```

Run against `target/release/jolt`, not source mode.

Also confirmed by hand in two unrelated projects: one with 40-odd tasks,
several colon-named, one carrying `:override-builtin`; and one with a plain
seven-task `bb.edn`. Both offer jolt's commands and their own tasks correctly.

### One note for anyone adding a jolt-core namespace

`jolt.completions` had to go into `host/chez/stdlib-fasl-manifest.txt`. Source
mode runs happily without the entry and only `make build` rejects it, so three
green gate runs said nothing about it.

Verified against `main` at 92fef2ce (v0.8.4) on 2026-09-07, macOS arm64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`jolt completions SHELL` prints a shell completion function**, for zsh, bash
+  or fish. `jolt <TAB>` then offers jolt's own commands and the project's tasks,
+  and under zsh each task carries its `:doc` as the description. Install it with
+  `source <(jolt completions zsh)` in `~/.zshrc` after `compinit`, or save the
+  output as `_jolt` on `$fpath`; both work.
+
+  The split from babashka is deliberate and is about latency. babashka calls its
+  binary back on every TAB press to compute the candidates for the line so far,
+  which it can afford. Jolt cannot: its floor is the Chez runtime coming up,
+  about 0.22s whatever it is asked for, and `jolt version` costs the same as
+  `jolt tasks`. So the two halves are split by how often they change. jolt's own
+  commands and options change when the binary does, so they are baked into the
+  snippet when it is generated and cost a completing shell nothing afterwards. A
+  project's tasks change when its `deps.edn` or `bb.edn` changes, so the snippet
+  caches them against those two mtimes and calls back only when one moves. The
+  zsh path uses `zsh/stat` and `$(<file)` and so forks nothing at all: a warm
+  press measures 0.4ms, against 97ms for the `bb tasks` that babashka's own
+  documented completion runs every time.
+
+  `jolt completions tasks` is the callback, and is useful on its own: one line
+  per listable task, `name<TAB>doc`, which is the machine-readable form of the
+  listing that `jolt tasks` writes for a person. Both go through
+  `jolt.tasks/listable`, so what TAB offers and what `jolt tasks` shows cannot
+  drift apart on which tasks exist. They differ deliberately on one point: a
+  task sharing a built-in's name is offered only when it wins that name with
+  `:override-builtin`, since a description says what the word will do and for a
+  task that loses to a command the answer is the command.
+
+  `make completionssmoke` gates the lines, all three snippets (parsed by their
+  own shells), the bash function run against a project, and the zsh function's
+  candidates read back through a stubbed `_describe` — that last one because
+  bash has no description column, so a candidate carrying the WRONG description
+  is invisible to every other check here.
+
 ### Fixed
 
 - **`jolt tasks` hides a task whose name starts with `-`.** `list-tasks!` is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **`jolt tasks` hides a task whose name starts with `-`.** `list-tasks!` is
+  babashka's listing, and babashka treats a leading dash the way it treats
+  `:private`: a helper another task calls, not an entry point someone picks off
+  a list. Jolt read the `:private` key and not the name, so a `bb.edn` using the
+  dash convention had its helpers listed. That file is one jolt reads directly,
+  so the convention arrives whether or not a jolt project would have chosen it.
+  Hiding is display only, here as in babashka: `jolt -dash` still runs the task.
+
+- **`jolt tasks` prints only the first line of a `:doc`.** The listing puts one
+  task on one line and aligns the docs into a column, so a docstring that spans
+  lines broke the shape it was being formatted into: the second line started at
+  column zero, in the name column, and read as a task of its own. Anything
+  parsing the listing for names picked it up as one. Babashka truncates for the
+  same reason.
+
 ## [0.8.4] - 2026-09-06
 
 A jolt file with `#!/usr/bin/env jolt` on its first line is an executable script:
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ install: build
 # naming the covered tree is written ONLY on a complete pass. `make gate-status`
 # answers "is this working tree gated?" — which is not something to remember.
 
-CI-GATES := submodules values corpus unit documented grenadine mvnhttp readscaling vecscaling pipescaling chunkscaling printscaling complexity ioscaling hotscaling depssmoke taskssmoke scriptsmoke depscpcache depsunit \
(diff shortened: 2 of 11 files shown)


yogthos:
Merged, along with #875. The latency argument is the right one — a per-press callback at a ~0.22s floor is not completion, and splitting the two halves by how often they change is the answer. Thanks also for the three zsh findings; `_describe` not keeping the first of a repeated value is the kind of thing only a real shell tells you, and the gate group you added for it is why it stays fixed.

On your design question — **keep it as a command.** `share/` would need install-path handling that differs between Homebrew, the curl installer, and a checkout, and would go stale against the binary that generated it. `jolt completions zsh` is the same one line everywhere.

Two follow-ups in #880, both from the code being written on macOS:

- The task cache never hit on Linux. Both snippets tried `stat -f %m` before `stat -c %Y`, and on GNU coreutils `-f` is a *filesystem* query — `%m` is read as a filename, a block table goes to **stdout**, and it exits 1. Only stderr was redirected, so the stamp ended up holding free-block counts, which move constantly. Measured: three presses gave three spawns, with `  File: "bb.edn"` as the stamp on disk. So the design worked and the spelling defeated it — invisibly, because the answers stayed right and only the latency came back.
- Fish cached one project's tasks for every other. The function stays loaded for the session and `__jolt_tasks_cache` had no key, so after `cd` TAB kept offering the first project's tasks. It also called jolt back outside a project entirely. Now keyed on the directory and gated on a config file. Fish was also passed `_opts` and ignored it, so `jolt -<TAB>` offered nothing there.

The gate now counts spawns across repeated presses rather than reading the snippet's text, since neither of those is visible to a grep.

Both ride in 0.8.5.

PR #891, opened 2026-09-08
Provide configurable tree shaking

### LINKED ISSUE #890: Add configurable tree shake
This asks for a declaration, in a library's or app's `deps.edn` that names lookup sites the author vouches for, so that tree shaking proceeds past them without keeping anything extra:

```clojure
:jolt/tree-shake {:allow-dynamic [clojure.spec.alpha/res
                                  clojure.spec.gen.alpha/dynaload]}
```

I ran into this using `clojure.spec.alpha` in my iOS app. 

```clojure
(ns hello.core (:require [clojure.spec.alpha :as s]))
(s/def ::n int?)
(defn -main [& _] (println (s/valid? ::n 1)))
```

```sh
$ jolt build -m hello.core -o hello --tree-shake
jolt build: tree-shake skipped (reachable code resolves vars at runtime):
  clojure.spec.gen.alpha/dynaload -> clojure.core/resolve
  clojure.spec.alpha/res -> clojure.core/resolve
```

With the declaration above in the app's `deps.edn`, the tree shake should run, and the hello world example should shrink by at least 9.7MB. EDIT: Turns out it was 13.3 MB smaller!

I note that GraalVM Native Image does exactly this with `reachability-metadata.json` under `META-INF/native-image/<group>/<artifact>/`: a library declares the dynamic access it needs, the app adds its own, and a shared repository covers libraries that have not.

statonjr:
`deps.edn` can now vouch for a runtime var lookup so `jolt build --tree-shake` proceeds past it:

```clojure
:jolt/tree-shake {:allow-dynamic [clojure.spec.alpha/res
                                  clojure.spec.gen.alpha/dynaload]}
```

The key is read from the app's `deps.edn` and every library's, where `:jolt/native` is read, and unioned. A `def` in the set is skipped by the bail scan. Nothing is kept on its behalf. With no key, the shake bails exactly as before.

When the shake still bails, the message lists each site once and ends with the paste-ready key:

```
jolt build: tree-shake skipped (reachable code resolves vars at runtime):
  clojure.spec.alpha/res -> clojure.core/resolve
to proceed, if these never run in the built binary, add to deps.edn:
  :jolt/tree-shake {:allow-dynamic [clojure.spec.alpha/res]}
```

Real-world check, the hello world from the issue with `spec.alpha 0.5.238`: bails with the two sites the issue names; with the pasted key it shakes, drops the compiler image, and `s/describe` still prints `int?`. Binary 30.7MB -> 17.4MB (13.3MB, 43%).

Fixes #890

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   adds one caller nothing vouches for and must still bail with a hint naming
   that caller alone. Verified by mutation: dropping either declaration bails
   the shaking fixture. `run-dce-refs.ss` pins the semantics on a synthetic
-  graph, including that an unreachable allowed def is still pruned.
+  graph, including that an unreachable allowed def is still pruned. (#890)
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **A `deps.edn` can vouch for a runtime var lookup, so `--tree-shake` proceeds
+  past it.** The shake bails, and must, when reachable code resolves a var by
+  name: the static graph cannot follow a runtime name. But some of those sites
+  are dead in a built binary and the author can prove it, and one `s/def` was
+  enough to lose the whole shake — spec.alpha's `res` only qualifies a symbol
+  so a spec's form can be printed, and spec.gen's `dynaload` sits behind a
+  `delay` nothing forces. `:jolt/tree-shake {:allow-dynamic [ns/name …]}` in
+  the app's `deps.edn` or any library's names the defs the author vouches for,
+  read where `:jolt/native` is read and unioned, so a library ships its list
+  once for every app that uses it. Nothing is kept on an allowed def's behalf.
+  With no declaration the shake bails exactly as before.
+
+  An allowed def is skipped by the compiler-needed scan as well as the bail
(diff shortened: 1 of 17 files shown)


yogthos:
Merged as 4bd11180 via #923. All eight of your commits are on `main` unchanged, with your authorship.

The detour: `main` had its history rewritten shortly before this landed, and this branch had to be rebased over that plus a `CHANGELOG.md` conflict. Updating the branch on your fork from here didn't work out, so the rebase went to a branch in this repo instead — hence the second PR number rather than this one showing as merged.

One review commit on top, both halves measured on the built binary rather than reasoned about:

- The failure mode wasn't written down anywhere. Vouch wrongly and the build still succeeds; the failure moves into the binary. A `resolve` of a def the shake dropped answers `nil` where the unshaken binary answers the var — silently — and an `eval` raises `direct-linked seed var clojure.core/eval is not bound to a procedure`, because the compiler image went with the same vouch. That's in the README next to the key now, and in the changelog entry.
- `run_local_case`'s new `EXPECT_OUT` read `$7` where the line beside it reads `${6:-default}`: harmless today, an unbound-variable error the day the script grows `set -u` the way `tasks-smoke.sh` has.

I reproduced #890 end to end against `spec.alpha 0.5.238`: the bail names both sites and prints the paste-ready key, pasting it shakes (358 of 828 defs kept), and `s/valid?` and `s/describe` answer identically. 30,900,404 → 17,582,651 bytes — your 13.3 MB / 43%, confirmed.

The design is the part worth calling out: reading the key where `:jolt/native` is read so a library ships its list once, printing the paste-ready line from the actual bail so the name survives inlining changes, and skipping the compiler-image check on the same vouch. The `^:redef` notes in the fixtures explaining why the bail names the def it does are exactly the right thing to have written down. Thanks.

PR #923, opened 2026-09-10
Provide configurable tree shaking (#891 rebased)

### LINKED ISSUE #890: Add configurable tree shake

yogthos:
#891 by @statonjr, rebased onto main and merged from this repo because main was rewritten under it and its fork branch could not be updated from here. Every commit but the last is his, unchanged.

Fixes #890.

`deps.edn` can vouch for a runtime var lookup so `jolt build --tree-shake` proceeds past it:

```clojure
:jolt/tree-shake {:allow-dynamic [clojure.spec.alpha/res
                                  clojure.spec.gen.alpha/dynaload]}
```

The key is read from the app's `deps.edn` and every library's, where `:jolt/native` is read, and unioned. A `def` in the set is skipped by the bail scan. Nothing is kept on its behalf. With no key, the shake bails exactly as before. When it still bails, the message lists each site once and ends with the paste-ready key.

## What the review added

One commit on top, both halves measured on the built binary:

- **The failure mode was not written down.** Vouch wrongly and the build still succeeds — the failure moves into the binary. A `resolve` of a def the shake dropped answers `nil` where the unshaken binary answers the var, silently; an `eval` raises (`direct-linked seed var clojure.core/eval is not bound to a procedure`) because the compiler image went with the same vouch. That is now in the README next to the key, and in the changelog entry.
- `run_local_case`'s new `EXPECT_OUT` read `$7` where its neighbour reads `${6:-default}` — harmless today, an unbound-variable error the day the script grows `set -u` as `tasks-smoke.sh` has.

## Verification

Reproduced the issue end to end against `spec.alpha 0.5.238` on this branch: the bail names both sites and prints the key, pasting it shakes (358 of 828 defs kept), and `s/valid?` and `s/describe` answer identically. 30,900,404 → 17,582,651 bytes, 13.3 MB / 43% — the reported figures.

`make test` green (109 CI targets, including `dcerefs` 77/77, `depsunit` 79/79 and `shakelocal` with both new fixtures). `make libconformance`: 46 ok, 0 regressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the app's `deps.edn` or any library's names the defs the author vouches for,
   read where `:jolt/native` is read and unioned, so a library ships its list
   once for every app that uses it. Nothing is kept on an allowed def's behalf.
-  With no declaration the shake bails exactly as before.
+  With no declaration the shake bails exactly as before. Vouching wrongly moves
+  the failure into the binary rather than failing the build: a `resolve` of a
+  dropped def answers `nil` there, silently, and an `eval` raises because the
+  compiler image went with the same vouch.
 
   An allowed def is skipped by the compiler-needed scan as well as the bail
   scan, on purpose: a site vouched never to run needs no compiler, and skipping
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **A `deps.edn` can vouch for a runtime var lookup, so `--tree-shake` proceeds
+  past it.** The shake bails, and must, when reachable code resolves a var by
+  name: the static graph cannot follow a runtime name. But some of those sites
+  are dead in a built binary and the author can prove it, and one `s/def` was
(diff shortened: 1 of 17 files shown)


sundbp:
This will be really useful!

yogthos:
yeah definitely a new knob to be able to tune ;)

PR #929, opened 2026-09-10
The reference types answer their declared interfaces through interop

yogthos:
Every watchable reference type was already an `IRef` *by class* — `(instance? clojure.lang.IRef (atom 1))` is true and `(supers clojure.lang.Atom)` lists `ARef` — but almost none of the methods those interfaces declare were answered. A library that reaches a reference through interop rather than through `clojure.core` got:

```
No matching method addWatch found taking 2 args for class clojure.lang.Atom
No matching field found: deref for class clojure.lang.Atom
```

`Var` was the only reference type with any method surface at all.

## What now answers

**`ARef` — the watch and validator half**, on all four of atom, ref, var and agent: `.addWatch`, `.removeWatch`, `.getWatches`, `.notifyWatches`, `.setValidator`, `.getValidator`.

There is no `clojure.lang.IWatchable` to implement — that is a ClojureScript protocol. On the JVM `getWatches`/`addWatch`/`removeWatch` are declared on `IRef` and implemented in `ARef`, which is why **one** arm serves all four types, and why a `Volatile` (an `IDeref`, not an `IRef`) correctly still has no `addWatch`.

**`IDeref` — the root every reference shares**: `.deref` and `.get` on atom, ref, agent, volatile, delay, promise, future and `reduced` (`get()` is a *default* method on `IDeref`, so every one of them carries it), plus the `.getAsBoolean` / `.getAsInt` / `.getAsLong` / `.getAsDouble` bridges — each the same cast `boolean`, `int`, `long` and `double` apply, so `(.getAsInt (atom 3.9))` truncates to `3`.

The two-argument `(.deref p 50 :timeout)` is `IBlockingDeref`'s and stays limited to a promise and a future. That arity is screened *before* dispatch on purpose: `jolt-deref` accepts the optional arguments and refuses them per type with the `ClassCastException` `@` raises, but the JVM never gets that far — there is no two-argument `deref` on a `Delay` to reflect onto — so `(.deref (delay 1) 50 :to)` has to be `No matching method`, not a cast error.

**Each type's own interface:**

| | methods |
|---|---|
| atom (`IAtom`/`IAtom2`) | `.swap` and `.swapVals` at all four arities incl. the JVM's `(f x y args)` spread form, `.reset`, `.resetVals`, `.compareAndSet` |
| ref | `.set`, `.alter`, `.commute` (the JVM's `(fn, args-seq)` shape), `.touch`, `.getMinHistory` / `.setMinHistory` / `.getMaxHistory` / `.setMaxHistory` / `.getHistoryCount` / `.trimHistory` |
| agent | `.getError`, `.getErrorMode`, `.setErrorMode`, `.getErrorHandler`, `.setErrorHandler`, `.getQueueCount`, `.restart`, `.dispatch` |

Every body is the native the `clojure.core` fn of that meaning is bound to, so a watch registered through interop and one registered through `add-watch` are **one list**, and a mutation through `.swap` or `.set` validates and notifies watches exactly as `swap!` and `ref-set` do.

## Four return values that were wrong

Wiring those natives turned up four whose `clojure.core` door answered the wrong thing — all handed back the agent, which threads but is not the value:

| | was | now (JVM) |
|---|---|---|
| `(restart-agent a 5)` | the agent | `5` — `Agent.restart` answers the new state |
| `(clear-agent-errors a)` | the agent | the state it restarted with |
| `(set-error-mode! a :continue)` | the agent | `nil` — `setErrorMode` is void |
| `(set-error-handler! a f)` | the agent | `nil` — `setErrorHandler` is void |

Fixed in the natives rather than papered over in the dispatch arm, so both doors agree. Nothing in `jolt-core` or `stdlib` read those returns.

## Where the code lives

Each table is keyed on name **and arity**, so a short call falls through to `dispatch-miss` and reports the JVM's `No matching method` instead of faulting inside a native on a short `rest`. The receiver guard runs before the name lookup, which is what leaves a `deftype` that declares these methods itself on its own dispatch.

`Agent`'s half is a registered method arm in `host/chez/java/concurrency.ss` rather than a clause in `records-dispatch.ss`'s base: its natives are there, that file loads long after `records-dispatch`, and the Gambit host does not include it at all — there is no agent there to have methods. The same reason splits `rd-deref-kind` across a hook, the shape `rd-class-method-hook` already uses. `getWatches` and `notifyWatches` have no `clojure.core` door at all, so they sit in `atoms.ss` beside the seam they read, which both hosts share.

## Verification

- **58 corpus rows** (21 `ARef` + 37 for the rest), **every one certified against reference JVM Clojure 1.12.5** — `make certify` reports **0 NEW divergences**, certified count 5465 → 5505.
- **3 unit rows** for `.dispatch`, which is the one method here that is a jolt *superset* rather than a match and so is deliberately not a corpus row: one serialized worker runs per agent, so there is no pool for the executor argument to select and it is accepted and ignored (the JVM, given a null executor, answers the agent and silently never runs the action). jolt has no `Agent/pooledExecutor` to name, so no common row exists.
- `host/gambit/records-gambit.ss` regenerated; `gambitgencheck` current.
- **Full `make test` gate passes** — 109 CI targets plus the self-host fixpoint — and `make gate-status` confirms the receipt matches the tree.


yogthos:
## CI failure fixed, and `jolt-8oa` folded in

### The CI failure was a racy corpus row of mine

```
=== NEW divergences (not in allowlist) — gate FAILS ===
  [iref / Agent interop methods] agent .getQueueCount
      actual: (let [ag (agent 0)] (await ag) (.getQueueCount ag))
      jolt-expected=0 JVM-result=1
```

Real race, not a semantic difference: the JVM's `await` sends a *wait action* to the agent and blocks on a latch, so it can return with that action still on the queue and `getQueueCount` answers 1 under load. It answered 0 on every local run, including five runs under full-gate CPU load — CI is the only place it showed, which is exactly why it had to go rather than be re-tried.

The portable row is now a fresh agent that was never sent to (`(.getQueueCount (agent 0))` → `0`), the only queue depth that is not a race to observe. The awaited form stays as a **unit** row, because it *is* deterministic on jolt — `await` there drains the queue and clears `running?` before returning.

Three neighbouring rows had the same shape of latent flake: they waited on a deliberately-failed agent with a fixed `(Thread/sleep 400)`. Those now poll `agent-error` to a deadline. Verified stable across five JVM runs under concurrent full-gate load.

### `jolt-8oa`: a `java.util` mutator name is not a method until its arity matches

An immutable collection is right to refuse a mutator it *has* with `UnsupportedOperationException`, but a bare list of names refused every one it merely **spelled**, at any arity. `java.util.List/set` is `set(int,E)`, so a one-argument `(.set [1 2] 9)` matches nothing on the JVM and is its `IllegalArgumentException` `No matching method set found taking 1 args for class clojure.lang.PersistentVector`. Refusing it instead both named a method the class does not have and put the call out of reach of the `(catch IllegalArgumentException …)` a caller writes. A no-argument `.add` and a one-argument `.clear` were the same mistake.

The interfaces also differ by receiver, and one predicate for both was answering `List`'s methods on a set — `(.getFirst #{1 2})` read `1` and `(.reversed #{1 2})` a reversed seq, where a `PersistentHashSet` is a `java.util.Set`, not a `SequencedCollection`, and the JVM has neither:

| receiver | surface |
|---|---|
| vector / list / seq / lazy seq | `java.util.List` + `SequencedCollection`: the positional overloads (`add/2`, `addAll/2`, `set/2`), the four ends (`addFirst`, `addLast`, `removeFirst`, `removeLast`), List's bulk ops (`replaceAll`, `sort`) |
| set, sorted or not | `Collection`'s surface and nothing more |

A **sorted** set turned out to be reaching none of this: it is an htable rather than a set record inside jolt, so every `Collection` mutator on one fell through to a miss instead of being refused. It now answers exactly as a hash set does — and a `PersistentTreeSet` is a plain `java.util.Set` on the JVM too, not a `SequencedSet`, so the Set half is the right half.

And on an **empty** collection three of them never reach a mutation to refuse, because they are default methods that walk the elements first: `.removeFirst` / `.removeLast` raise `NoSuchElementException` (the check `.getFirst` / `.getLast` already made), `.removeIf` answers `false`, and `.sort` / `.replaceAll` are void and do nothing.

**280 name × arity × receiver cells** checked against reference Clojure. 278 match; 18 corpus rows pin the distinct rules.

The two that jolt deliberately does not match are worth naming, because the cause is upstream: `LazySeq.reify()` returns a **mutable** `new ArrayList(this)` where `ASeq.reify()` wraps it in `Collections.unmodifiableList`. `List`'s default `sort` and `replaceAll` both go through `listIterator()`, so on a lazy seq they mutate a throwaway copy and silently succeed, while refusing on every other seq. jolt refuses uniformly — reproducing it would mean reproducing a missing wrapper whose only observable is that nothing happens. Recorded as `:upstream-quirk` in `known-divergences.edn`.

### One thing found on the way, filed not fixed

The first local `make test` after this change reported 6 corpus divergences and 5 crashes for code that a standalone `make corpus` passed twice, before and after. That was a **stale boot image**, and the mechanism is a pre-existing race worth flagging:

`make-gateboot.ss` renames `target/dev/gate.inputs` into place at step 1, then spends ~10s compiling and renames `gate.so` at step 2. `gate-boot-fresh.ss` decides freshness by comparing the hashes *in* `gate.inputs` against current source — so inside that window the list already matches the tree while the image on disk is the previous build. Under `make -j ci` a concurrent gate target checks freshness there, is told `#t`, and loads the stale image. File times confirmed it: `gate.inputs` 15:06:32, `gate.so` 15:06:42, with the 22s corpus run starting in the gap.

That is the false-green `gate-boot-fresh.ss`'s own header is about, reached by a torn write of the `(inputs, so)` pair rather than by mtime. The per-file `.tmp`+rename is already there; what is missing is ordering *across* the pair — writing `gate.inputs` only after `gate.so` lands would make the window fail closed. Filed as a P1 with the repro rather than fixed here, since it is infrastructure and unrelated to this PR; `make-devboot.ss` has the same `flat.inputs` / `flat.so` shape and wants checking too.

### Gate

Full `make test` green on this exact tree (109 CI targets + self-host fixpoint, `make gate-status` confirms the receipt). `make certify`: **0 NEW, 0 stale**, 111 of 111 divergences known, certified 5465 → 5522.

PR #966, opened 2026-09-12
java.io.PushbackInputStream

### LINKED ISSUE #965: Add java.io.PushbackInputStream support
Please take a look [fix: support java.io.PushbackInputStream interop](https://github.com/jolt-lang/jolt/commit/8cb46e638f94c8ddf6a570f3757cc4d7b4733ef8)

I created it to use datalevin via. babashka.pods [dep.edn](https://github.com/dukefong8/Playground/blob/main/jlt/deps.edn). The test is in https://github.com/dukefong8/Playground/blob/main/jlt/test/babashka/pod/datalevin_test.clj

Thanks!

yogthos:
Fixes #965.

`(java.io.PushbackInputStream. in)` / `(java.io.PushbackInputStream. in size)` over any InputStream: jolt's own, a process or socket stream, or a `proxy`. `unread` takes an int, a `byte[]` or a range of one, the capacity is the JVM's fixed one ("Push back buffer is full" past it), `available` counts the pushed-back bytes on top of the wrapped stream's, `close` reaches the wrapped stream, `markSupported` is false. bencode reads one byte to classify a token and puts it back before reading a netstring, which is how babashka.pods and nrepl talk to a process; neither could load.

The approach differs from the commit linked in the issue. There the pushed-back bytes sat in a list beside the stream and `in-stream?` was widened to a second tag, so every site that reaches a stream's port directly (`slurp`, `io/copy`, `BufferedInputStream.`, `drain-any-stream`, read-line via `setIn`) had to be patched one by one, and any site not patched silently dropped the byte. It also required the wrapped value to be one of jolt's own streams, which rejects a proxy. Here a pushback stream is an ordinary in-stream whose port is a custom binary port reading through the wrapped stream's own `read`, and `unread` writes into the front of that port's buffer. The port is the one place every reader goes, so the pushed-back bytes come first for all of them by construction. `read(byte[] …)` keeps the JVM's rule that pushed-back bytes falling short of `len` are followed by one read of the wrapped stream; `read()` with nothing buffered is the wrapped stream's own `read()`, as on the JVM.

Second commit: a `proxy` over `java.io.InputStream` / `OutputStream` / `Reader` / `Writer` inherits the class's concrete methods. jolt generates no class, so `(proxy [java.io.InputStream] [] (read …))` had no `readAllBytes`, `skip`, `available`, `close`, … and `slurp`, `io/copy`, `spit`, `io/writer`, `with-open` and a PushbackInputStream over one all failed on the missing method. The four roots' inherited surfaces are method tables a reify's method miss consults through the classes it declares and their ancestry (`register-abstract-methods!` in host-static.ss, a hook in records-dispatch.ss), the method running against the reify so its calls back into `read`/`write` reach the override as the JVM's virtual dispatch does. An abstract method the body omits is the proxy macro's stub (`UnsupportedOperationException` naming it). The coercion sites treat such a value as the class it declares, and encoders hand an OutputStream `write(byte[],int,int)` as the JVM's do. Certifying the rows against the JVM corrected my model of Clojure's proxy: the fn stands in for every overload of a name, so there is no per-arity inheritance to model.

Found along the way and fixed: the concrete `java.io` stream classes had constructors but no class-graph rows (`isa?` false, `supers` nil); `(str baos)` rendered as `#object[…]` (transit clients read the payload with it); an OutputStream flushed or closed after close, or written to, raised a classless Chez port error instead of a no-op / `IOException`. The in-stream `instance?` collapse is a machine-checked known-divergences entry now. Filed jolt-idii: `str` of a jhost should be its `toString` generally, which needs an audit of the shim `toString` methods first.

Tests: 38 corpus rows certified against JVM Clojure 1.12.5 (io / PushbackInputStream, io / java.io stream class graph, interop / proxy over an abstract stream class, io / ByteArrayOutputStream, io / OutputStream after close), three system-streams smoke checks over real stdin and a `setIn`, an end-to-end babashka.pods 0.2.0 round trip against a pod written in jolt (both sides read through a PushbackInputStream, 200 invocations). `make ci` green, `make libconformance` 47 ok / 0 regressed.


yogthos:
Reviewed against #975, which implemented the same class from @dukefong8's original commit. Taking this one.

The deciding difference is where the pushed-back bytes live. #975 kept them in a prefix buffer beside the stream, so every site that reaches a stream's port directly had to be threaded through it; two were missed, and both fail silently rather than loudly:

```clojure
(let [p (java.io.PushbackInputStream. (java.io.ByteArrayInputStream. (.getBytes "bc")))]
  (.unread p 97)
  (String. (.readAllBytes (java.io.BufferedInputStream. p))))
;; JVM "abc"   #975 "bc"   this PR "abc"

(java.io.PushbackInputStream. (proxy [java.io.InputStream] [] (read ([] -1))))
;; #975 IllegalArgumentException; this PR fine
```

Putting them in the stream's own port buffer is what makes the enumeration unnecessary. I ran all nine coercion paths — `BufferedInputStream`, `io/reader`, `InputStreamReader`, `slurp`, `io/copy`, `readAllBytes`, `io/input-stream`, `line-seq`, and a `proxy` source — against reference Clojure 1.12.5 and this branch agrees on every one.

Merged `main` in (CHANGELOG only, both sides additive under `### Fixed`). Gates re-run on the merge: `unit` 1819/1819, `corpus` 5696/5706 with 0 new divergences, `documented` 57/57, `systemstreams` 51/51.

#975 is now the BigDecimal half alone and closes #968 only; this PR closes #965.

PR #974, opened 2026-09-13
Close the three runtime gaps that keep SCI from evaluating protocol code (#969, #970, #971)

### LINKED ISSUE #969: `apply` ignores `IFn.applyTo` on deftype/record/reify callables, breaking >20-arg application
**Component:** runtime (host/chez) — `apply` / IFn dispatch
**Version:** `jolt v0.8.7-7-gc6a3878d`, linux (Termux/aarch64), release build
**Severity:** high — any library that constructs `clojure.lang.IFn`-implementing values and applies them to >20 args breaks; in particular **SCI cannot evaluate a call to a core function with more than 21 arguments**, so `(str …)` / `(vector …)` / any variadic core fn with a long literal arg list throws at runtime.

## Summary

`apply` routes everything through `jolt-invoke`, which for a `deftype`/`defrecord`/`reify` implementing `clojure.lang.IFn` dispatches to the type's `invoke` method by arg count. `IFn.applyTo` is never consulted, and the 21-parameter `invoke` (the JVM varargs arity, whose last parameter receives the rest args) is treated as a fixed 21-arg method. Host-callable fns (`AFn`/`RestFn`, handled via `jolt-register-variadic!`) are fine; only non-`AFn` IFn implementers are affected.

## Minimal reproduction (no deps, no SCI)

```clojure
(defrecord R []
  clojure.lang.IFn
  (invoke [this] :zero)
  (invoke [this a] a)
  (invoke [this a b] [a b])
  (invoke [this a b c d e f g h i j k l m n o p q r s t] :twenty)
  (invoke [this a b c d e f g h i j k l m n o p q r s t rest] (count rest))
  (applyTo [this args] (count args)))

(def r (->R))
(doseq [n [1 20 21 22 30]]
  (println "n=" n "=>" (try (apply r (range n))
                           (catch Throwable e (str (class e) ": " (ex-message e))))))
```

### Expected (JVM Clojure 1.11 — `apply` calls `applyTo`)

```
n= 1 => 1
n= 20 => 20
n= 21 => 21
n= 22 => 22
n= 30 => 30
```

### Actual (jolt)

```
n= 1  => 0
n= 20 => :twenty
n= 21 => class java.lang.UnsupportedOperationException: count not supported on this type: Long
n= 22 => class clojure.lang.ArityException: Wrong number of args (23) passed to: fn
n= 30 => class clojure.lang.ArityException: Wrong number of args (31) passed to: fn
```

The `n=21` case shows the last element being fed into the varargs `rest` slot (a `Long`, then `count`ed); `n≥22` shows an `ArityException` with `n+1` (the leading `this`).

## Root cause

`host/chez/seq.ss`:

* `jolt-apply` (line **1924**) special-cases only `jolt-concat` and fns registered via `variadic-fixed-arity-of` (host `AFn`s). Every other callable is sent to `jolt-invoke` — `IFn.applyTo` is never used:

  ```scheme
  (else (apply jolt-invoke f (append fixed (seq->list (jolt-seq tail)))))
  ```

* `jolt-invoke` (line **1230**) handles record/reify IFn implementers by *only* looking for an `invoke` method and re-dispatching by arg count (lines **1263–1266**):

  ```scheme
  ((and (jrec? f) (find-method-any-protocol (jrec-tag f) "invoke"))
   => (lambda (m) (apply jolt-invoke m f args)))
  ((and (reified-methods f) (hashtable-ref (reified-methods f) "invoke" #f))
   => (lambda (m) (apply jolt-invoke m f args)))
  ```

  There is no equivalent arm for `applyTo`, and no handling of the JVM convention that an `invoke` with 21 params (after `this`) is the varargs entry point.

## Real-world impact (how this was found)

`org.babashka/sci`'s `sci.lang.Var` (`vendor/sci/src/sci/lang.cljc`, line **178**) is exactly such a value: it implements `clojure.lang.IFn` with fixed `invoke` arities 0–20 plus the 21-param varargs method and `applyTo` (line **224**). SCI compiles calls with ≥3 args to `(apply f args)` where `f` is that `sci.lang.Var`. So evaluating

```clojure
(sci/eval-string* (sci/init {}) "(str \"a\" …22 args…)")
;; => wrong number of args (23) passed to: fn     (21 args still works; 22+ fails)
```

Both `org.babashka/sci 0.13.53` (Maven) and `~/jolt/vendor/sci` fail identically, so it is not an SCI-version problem. In kmet this aborts loading an extension whose source contains a `(str …35 args…)` form.

## Suggested fix

In `jolt-apply`, before falling through to `jolt-invoke`, consult the callable's `applyTo` when it is not a host procedure:

* if `f` is a record/reify with an `applyTo` method, invoke it with the tail as a single seq argument (JVM `IFn.applyTo` semantics); or
* more precisely mirror the JVM: for a non-`AFn` IFn implementer, `apply` is `(.applyTo f (seq (concat fixed tail)))`, and only the *type's own* `invoke`/varargs method should be selected when `applyTo` is absent.

(Alternatively, treat the 21-param `invoke` as the varargs arity so `(apply f a…a20 rest-seq)` reaches it correctly.)

## Regression test to add

The repro above belongs in jolt's test suite: `(apply (->R) (range n))` must equal `n` for `n` in `{1,20,21,22,30}`, asserting `applyTo` is the entry point for non-`AFn` IFn implementers.

### LINKED ISSUE #970: `clojure.lang.MultiFn` has no JVM-compatible constructor — `new`/`host-new` throws "No matching ctor found"
**Component:** runtime (host/chez) — `host-new` / ctor table; multimethods
**Version:** `jolt v0.8.7-7-gc6a3878d`, linux (Termux/aarch64), release build
**Severity:** high — `defprotocol` is unusable under any SCI-based evaluation on jolt (SCI's `defprotocol` expands to `defmulti`, which constructs a `clojure.lang.MultiFn` via `new`). This blocks every `defrecord`/`deftype` that implements a protocol inside SCI — including the entire kmet extension system, whose components are `defrecord`s implementing `IComponent`.

## Summary

jolt's multimethods are the custom `jolt-multifn` record; the runtime's own compiler (`defmulti`/`defprotocol` macros) builds them through `make-jolt-multifn`. But there is no constructor registered for the **class name** `clojure.lang.MultiFn`, so the JVM reflection spelling `(new clojure.lang.MultiFn name dispatch default hierarchy)` — which arbitrary Clojure code (notably SCI) emits — falls through `host-new`'s ctor lookup to the "no ctor" throw.

## Minimal reproduction

```clojure
;; both spellings fail
(clojure.lang.MultiFn. "m" identity :default nil)
(clojure.core/host-new "clojure.lang.MultiFn" "m" identity :default nil)
```

* **JVM Clojure 1.11:** `ctor ok: clojure.lang.MultiFn true` (the public 4-arg ctor `MultiFn(String, IFn, Object, IPersistentMap)` exists)
* **jolt:**
  ```
  Unhandled exception (IllegalArgumentException): No matching ctor found for class clojure.lang.MultiFn
  ```

Note `defmulti` and `defprotocol` written *in plain jolt* work fine — the compiler emits `clojure.core/defmulti-setup` / `make-protocol`, which bypass this path. Only reflective/`new` construction is missing.

## Real-world impact

`org.babashka/sci`'s `defprotocol` expands to `defmulti`, whose implementation is `sci.impl.multimethods/multi-fn-impl`:

```clojure
(new clojure.lang.MultiFn name dispatch-fn default hierarchy)
```

So a bare SCI context cannot define a protocol at all:

```clojure
(sci/eval-string* (sci/init {}) "(do (defprotocol P (m [x])) (defrecord R [] P (m [x] :ok)) (m (->R)))")
;; => No matching ctor found for class clojure.lang.MultiFn
```

Both `org.babashka/sci 0.13.53` (Maven) and `~/jolt/vendor/sci` fail identically, so it is a jolt runtime gap, not an SCI version issue. In kmet this surfaces while loading any extension that uses `kmet.tui.macros/defcomponent` (a `defrecord` implementing `kmet.tui.protocols/IComponent`).

## Root cause

`host/chez/java/host-static.ss` `host-new` (line **865**) looks the class up in `class-ctors-tbl`; `clojure.lang.MultiFn` has no entry, `lib-try-autoload!` finds no provider, and the deftype/record var fallback finds no `clojure.lang.MultiFn` var, so it throws `No matching ctor found for class clojure.lang.MultiFn`.

jolt already knows how to build the value: `make-jolt-multifn` in `host/chez/multimethods.ss` (used by `jolt-defmulti-setup`, line **115**), and MultiFn interop (`.addMethod`, `.getMethod`, `.preferMethod`, `.removeMethod`, `.methods`) is already wired for defmulti values in `host/chez/records-dispatch.ss` (line **310**).

## Suggested fix

Register a `clojure.lang.MultiFn` constructor (FQ and short name) in the ctor table, matching the JVM signature `(String name, IFn dispatchFn, Object defaultVal, IPersistentMap hierarchy)`, implemented by factoring the construction out of `jolt-defmulti-setup`:

```scheme
(make-jolt-multifn name-string dispatch default
                   (new-mm-table) default hierarchy (new-mm-table) (new-mm-table) -1 #f)
```

(`nil` hierarchy should behave as the global hierarchy, as on the JVM.) A provider could also register it via `clojure.core/__register-class-ctor!`, but since the class is runtime-implemented this belongs in the runtime.

## Regression test to add

1. `(clojure.lang.MultiFn. "m" identity :default nil)` returns a value that is `instance? clojure.lang.MultiFn`, and `(.addMethod m :x (fn [& _] :x))` + dispatch through it works.
2. SCI end-to-end: `(defprotocol P (m [x]))` no longer throws.

> **Follow-up (separate issue):** with the ctor shimmed in at runtime, `defprotocol` then succeeds but the same SCI eval fails later at `(defrecord R [] P …)` with `Protocol not found: P` — i.e. SCI's `deftype`/`defrecord` protocol lookup (`eval-resolve-state` → `resolve/lookup`) does not resolve the protocol under jolt (plain `(resolve 'clojure.core/inc)` also returns `nil` inside a jolt SCI context). Worth filing/verifying separately once this ctor lands.

### LINKED ISSUE #971: `extend-protocol` onto a deftype/record mutates the class, so `instance?` against the protocol interface is true (JVM `extend` leaves the class untouched)
**Component:** runtime — protocols / `extend` / `extend-protocol` / type metadata
**Version:** `jolt v0.8.7-7-gc6a3878d`, linux (Termux/aarch64), release build
**Severity:** high — this is why `org.babashka/sci` cannot `resolve` / define records under jolt. SCI's `eval-node?` uses `(instance? <protocol-interface> x)` and relies on the JVM rule that `extend` never modifies the class. On jolt it does, so SCI misclassifies its own `sci.lang.Var`, and every `defrecord`/`deftype` implementing a protocol fails with `Protocol not found`.

## Summary

`extend-protocol` (and presumably `extend`) on a jolt-defined `deftype`/`defrecord` attaches the protocol methods — plus a `__jolt_extend__` helper — to the type's class. The methods then show up in reflection and make `(instance? <protocol-interface> x)` true for a type that only had the protocol *extended* onto it, never declared it.

On the JVM, `extend`/`extend-protocol` records the implementation in the protocol's method table only; the class is not touched. `satisfies?` becomes true, method dispatch works, but `instance?` of the protocol's interface stays false and the class gains no method.

## Minimal reproduction

```clojure
(defprotocol Pe (pe [x]))
(deftype D [])
(extend-protocol Pe D (pe [x] :d))

(sort (mapv #(.getName %) (.getDeclaredMethods (class (->D)))))
(satisfies? Pe (->D))
(pe (->D))
```

| | jolt | JVM Clojure 1.11 |
|---|---|---|
| `.getDeclaredMethods` | `(__jolt_extend__ pe)` | `(getBasis)` |
| `(satisfies? Pe (->D))` | `true` | `true` |
| `(pe (->D))` | `:d` | `:d` |

The behavioral difference is `instance?`: on jolt it is driven by the now-present `pe` method on the class (so it returns true); on the JVM the interface was never added, so `instance?` of the protocol interface is false.

## Impact (how it was found)

SCI extends the `sci.impl.types/Eval` protocol onto concrete runtime types — including `sci.lang.Var` — and then uses `instance?` against it as a *direct-implementation* test:

```clojure
;; sci/impl/evaluator.cljc:336
(extend-protocol types/Eval
  java.lang.Class            (eval [expr _ _] expr)
  clojure.lang.PersistentArrayMap (eval [expr _ _] expr)
  clojure.lang.Symbol        (eval [expr _ _] expr)
  sci.lang.Namespace         (eval [expr _ _] expr)
  sci.lang.Var               (eval [expr _ _] expr)   ; <-- extended, not implemented
  ...
  Object                     (eval [expr _ _] expr))

;; sci/impl/types.cljc:154
(defn eval-node? [x]
  (instance? sci.impl.types.Eval x))
```

`eval-node?` is meant to be true only for the AST node types that *directly* implement `Eval` (`Node`, `BindingNode`, …); the `extend-protocol` entries exist for the `eval` function and must not flip `instance?`. On jolt they do:

```
(instance? sci.impl.types.Eval <a sci.lang.Var>)   ; jolt: true (should be false)
```

`eval-resolve` then discards every resolved value:

```clojure
(when-not (sci.impl.types/eval-node? res) res)     ; res is the var -> nil
```

which cascades into:

* `(resolve 'clojure.core/inc)` → `nil` inside a jolt SCI context (should be `#'clojure.core/inc`);
* `ns-resolve` likewise;
* `deftype`/`defrecord` protocol lookup (`standard-scitype-path` calls `eval-resolve-state` for the protocol symbol) fails:

```clojure
(sci/eval-string* (sci/init {}) "(do (defprotocol P (m [x])) (defrecord R [] P (m [x] :ok)) (m (->R)))")
;; => Protocol not found: P
```

So no protocol-based code evaluates under SCI on jolt — e.g. kmet extensions, whose components are `defrecord`s implementing `kmet.tui.protocols/IComponent`, cannot load at all. (Plain jolt `defprotocol`/`defrecord` is fine; only the SCI path is affected.)

## Root cause / suggested fix

jolt's `extend`/`extend-protocol` should not add the protocol methods (or `__jolt_extend__`) to the target type's class. Like the JVM, it should register the implementation only in the protocol's method/preference tables, leaving the class — and therefore `instance?` against the protocol interface and reflection — unchanged. There appears to be a fold-into-the-class optimization for jolt-defined types that must be removed, or at least must not affect `instance?`/`.getDeclaredMethods`.

## Regression test

1. `(defprotocol Pe (pe [x])) (deftype D []) (extend-protocol Pe D (pe [x] :d))`:
   * `.getDeclaredMethods` of `D` contains no `pe`/`__jolt_extend__`;
   * `(instance? <Pe interface> (->D))` is `false`;
   * `(satisfies? Pe (->D))` is `true` and `(pe (->D))` returns `:d`.
2. SCI end-to-end on jolt: `(resolve 'clojure.core/inc)` returns the var, and `(defprotocol P (m [x])) (defrecord R [] P (m [x] :ok)) (m (->R))` returns `:ok`.

yogthos:
Three independent runtime misses with one visible consequence: no SCI context on jolt could define a protocol, resolve a var, or call a core fn with more than 21 arguments. Each is a real gap on its own; they are in one PR because they sit one behind the other on the same path and the end-to-end check only passes with all three.

### `extend-protocol` made the target an `instance?` of the protocol's interface (#971)

On the JVM, `extend`/`extend-type`/`extend-protocol` file the implementation in the protocol's own method table and never touch the class: `satisfies?` and dispatch flip, `instance?` of the protocol's interface stays false, and the class gains no method. jolt files both kinds of implementation in one registry, and the extend mark it already writes for `extenders` is what tells them apart — so the `instance?` walk and `Class.getDeclaredMethods` now skip an extended protocol, and the `__jolt_extend__` marker stops appearing as a method of the class.

`sci.impl.types/eval-node?` is `(instance? sci.impl.types.Eval x)`, meant to be true only for the AST nodes that *declare* `Eval`; SCI extends the same protocol onto `sci.lang.Var` and friends for the `eval` function's sake. Answering true there made jolt classify every resolved value as an AST node, and `eval-resolve` discarded it — `(resolve 'clojure.core/inc)` was `nil`, and `defrecord` could not find its protocol ("Protocol not found: P").

### `clojure.lang.MultiFn` had no constructor (#970)

`defmulti` and `defprotocol` written in jolt expand to `defmulti-setup`/`make-protocol` and build the value directly, so the class the runtime models answered `(class x)` and its interop methods but not `new`. `sci.impl.multimethods/multi-fn-impl` is `(new clojure.lang.MultiFn name dispatch-fn default hierarchy)` and SCI's `defprotocol` expands to `defmulti`, so that spelling is not a corner case. The construction is factored out of `jolt-defmulti-setup` as `make-multifn` and registered as the ctor under both names. The hierarchy argument is an `IRef` on the JVM, deref'd per dispatch (SCI passes `#'clojure.core/global-hierarchy`), so the three places that read the field now unwrap a var as well as an atom.

### `apply` ignored `IFn.applyTo` (#969)

`jolt-apply` sent every non-procedure callable to `jolt-invoke`, which looks for `invoke` alone and selects it by argument count — so a 21-parameter varargs `invoke` was treated as a fixed 21-arg method and anything past it was an `ArityException`. On the JVM `apply` is `(.applyTo f (seq args))` for every `IFn` that is not an `AFn`, which is the only entry point such a type has for more arguments than it declares an arity for. A `deftype`/record/`reify` that declares `applyTo` now gets it, with the argument seq unrealized; anything without one keeps the `invoke` dispatch it had.

### Verification

End to end, on this branch, through the vendored SCI:

```clojure
(def ctx (sci/init {}))
(sci/eval-string* ctx "(resolve 'clojure.core/inc)")            ;=> #sci.lang.Var{…}   (was nil)
(sci/eval-string* ctx "(do (defprotocol P (m [x])) (defrecord R [] P (m [x] :ok)) (m (->R)))")
                                                                 ;=> :ok               (was "No matching ctor found for class clojure.lang.MultiFn")
(sci/eval-string* ctx "(str 0 1 2 … 29)")                        ;=> "0123…29"         (was ArityException at 22 args)
```

Gates run locally: `unit`, `corpus`, `documented`, `selfhost`, `sci`, `scifunctional`, `smoke`, `cts`, `gambitgencheck`, `gambitseedcheck`, `attributioncheck`.

### Tests

- `extend-protocol` leaving `instance?` false while `satisfies?` and dispatch flip — the row that pinned the old behaviour now pins this one, and a second row keeps the `instance?` memo's invalidation covered through an inline redefinition (which is what *does* add the interface, here as on the JVM).
- `applyTo` as `apply`'s entry point at 0/1/20/21/22/30 arguments, on a record and on a `reify`, plus a row asserting a direct call still selects `invoke`.
- The `MultiFn` ctor: dispatch, the `:default` value, and a var hierarchy resolving `isa?` dispatch.

Closes #969. Closes #970. Closes #971.


yogthos:
One latent bug found in review, fixed in `0c7256a`. Making `instance?` answer for the *declaration* also took the answer away from a type that declares the protocol inline and is *then* extended with it:

```clojure
(defprotocol P (m [_]))
(deftype T [] P (m [_] :inline))
(instance? user.P (->T))            ; true
(extend-type T P (m [_] :extended))
(instance? user.P (->T))            ; false on this branch, and satisfies? still true
```

`mark-extend!` writes into the impl table the inline methods already filled, so one mark could not tell the two apart. The JVM never has to — it refuses that extend outright (`class user.T already directly implements interface user.P for protocol`) — but the class question still has its answer: the class declares the interface, so an extend cannot take `instance?` away. `register-inline-protocol!` now writes an inline mark of its own and it wins; a table carrying only the extend mark is still the extend-only case #971 is about.

Both bookkeeping marks are filtered out of `Class.getDeclaredMethods`, which walks those tables by name — the inline mark would otherwise have shown up as a method the way `__jolt_extend__` used to.

Unit rows added for the declare-then-extend pair and for neither mark being reflected; `gambitgen` re-run. Gates on the merge with `main`: `unit` 1828/1828, `corpus` 5696/5706 with 0 new divergences, `gambitgencheck` current.

PR #977, opened 2026-09-13
Keep clojure.core.async out of all-ns until it is required (#972)

### LINKED ISSUE #972: `clojure.core.async` is interned at startup with a partial var set, so `all-ns`/`find-ns` expose an incomplete namespace
**Component:** runtime / namespaces (vendored stdlib)
**Version:** `jolt v0.8.7-7-gc6a3878d`, linux (Termux/aarch64), release build
**Severity:** medium — code that snapshots namespace contents from `all-ns` (interpreter contexts, doc/completion tooling) captures 34 of 130 vars and never sees the rest, even though a later `require` resolves the namespace fine.

## Summary

In a fresh jolt process `clojure.core.async` is already created and present in `(all-ns)` / `(find-ns)`, but interned with only a bootstrap subset of its vars. `(require 'clojure.core.async)` then interns the remaining ~96 into that same namespace object. Every other vendored built-in (`babashka.fs`, `babashka.process`, `clojure.core.rrb-vector`, …) is absent from `find-ns`/`all-ns` until required — `clojure.core.async` is the outlier.

## Reproduction

```clojure
(println (count (ns-interns (find-ns 'clojure.core.async))))    ; 34
(println (contains? (ns-interns 'clojure.core.async) 'alts!!))  ; false
(require 'clojure.core.async)
(println (count (ns-interns (find-ns 'clojure.core.async))))    ; 130
(println (contains? (ns-interns 'clojure.core.async) 'alts!!))  ; true
```

Present at startup (34): `<! <!! >! >!! chan chan? close! put! take! offer! poll! promise-chan buffer dropping-buffer sliding-buffer unblocking-buffer? timeout go go-loop go-spawn go-monitor thread thread-spawn fiber-execute fiber-spawn …`

Missing until required (96): `alts! alts!! to-chan to-chan! merge mix mult pub pipe pipeline pipeline-async map filter map< filter< reduce into onto-chan split tap sub unmix …`

Contrast, same fresh process:

```clojure
(find-ns 'babashka.fs)        ; nil
(find-ns 'babashka.process)   ; nil
(find-ns 'clojure.core.async) ; #namespace[...]   <- created, incomplete
```

## Impact (how it was found)

kmet's extension loader builds an SCI context by scanning `all-ns` and injecting the shared namespaces **by reference** (`sci/copy-var*` over `ns-interns`). It captured the 34-var `clojure.core.async` into the context, shadowing the real namespace: an extension's `(async/alts!! …)` then failed with `Unable to resolve symbol: async/alts!!` even though the `(:require [clojure.core.async :as async])` itself resolved — SCI was looking at the injected partial namespace. Any consumer that snapshots namespaces from `all-ns` before they are used hits the same truncation.

## Suggested fix

Do not create / intern `clojure.core.async` at startup (the bootstrap subset should live under a private name, or the namespace should not be listed in `all-ns` until it is genuinely loaded). A namespace should appear in `all-ns`/`find-ns` complete, exactly as the other vendored built-ins do.

yogthos:
In a fresh process `clojure.core.async` was already in `(all-ns)` and `(find-ns)`, interned with 34 of its 130 vars.

The channel primitives are native — `host/chez/java/async.ss`, `fibers-async.ss` and `sm.ss` `def-var!` `chan`, `<!`, `>!`, `go`, `thread` and the buffers into the namespace at boot — and the var-derived existence rule behind `find-ns` needs nothing more than that, so the namespace existed from the first instant, three-quarters empty. The other ~96 vars (`alts!`, `mult`, `mix`, `pub`/`sub`, `pipeline`, the sequence operators) arrive when a `require` pulls `stdlib/clojure/core/async.clj` off the source roots.

Reporting it that way is worse than not reporting it. A caller that *snapshots* namespaces out of `all-ns` — an SCI context built by copying `ns-interns`, a completion or doc index — captures the bootstrap subset and then stands in for the real namespace with it, while the same program's own `(require '[clojure.core.async :as async])` resolves perfectly well. The mismatch surfaces far from its cause, as `Unable to resolve symbol: async/alts!!` against a namespace that is plainly required. Every other vendored built-in (`babashka.fs`, `babashka.process`, `clojure.core.rrb-vector`) is simply absent from `find-ns` until required, and `clojure.core.async` was the outlier.

```clojure
;; before                                  ;; after
(find-ns 'clojure.core.async)  ;=> #ns[…]  ;=> nil
(count (ns-interns …))         ;=> 34
(require 'clojure.core.async)
(find-ns 'clojure.core.async)  ;=> #ns[…]  ;=> #ns[…]
(count (ns-interns …))         ;=> 130     ;=> 130
```

`intern-ns!` is what lifts the deferral, so the namespace appears the moment its own `ns` form runs — which is to say when the overlay actually loads — and `in-ns`/`create-ns` lift it too, as they would for any name.

**Scope:** only `find-ns` and `all-ns` consult the deferral. `ns-has-vars?` and `chez-ns-exists?` keep answering from the var table, so the loader's already-in-memory arm and the analyzer's "No such namespace" guard are unchanged, and a fully-qualified `clojure.core.async/chan` still compiles and runs without a `require`, as it has since the primitives became native. Hiding the namespace from the *enumeration* is the whole change.

### Tests

A fresh process is the only place this is visible, so the three cases are smoke cases: absent from `find-ns` and `all-ns` before the require, present and complete after it, and the primitives still resolving fully qualified without one.

Gates run locally: `unit`, `corpus`, `selfhost`, `smoke`.

Closes #972.


yogthos:
Merged. One thing found in review and fixed in `23e0ffd` before landing: the deferral leaked through anything that *describes* a var.

```clojure
(boolean (some #{'clojure.core.async} (map ns-name (all-ns))))  ;=> false
(pr-str (resolve 'clojure.core.async/chan))
(boolean (some #{'clojure.core.async} (map ns-name (all-ns))))  ;=> true, 34 vars
```

Two holes, both from treating `intern-ns!` as the declaration. It only materializes the jns *object* for a name, and the things that reach it are often just describing a var — `(.ns v)` and a var's `:ns` metadata both go through it, and printing a var goes through them. And `all-ns` unioned `ns-registry` unfiltered, so once an object existed for the name it was reported whatever the deferral said, with `find-ns` reading the registry before consulting it.

So the lift moved to what actually declares or loads the namespace (`ldr-mark-loaded!`, `in-ns`, `create-ns`), and the deferral now answers first in `find-ns` and filters both halves of `all-ns`. A smoke case pins it — a completion or doc index that resolves and prints is exactly the caller in the PR description.

PR #980, opened 2026-09-13
Block-fill the reader drain and recognize literal regex patterns

yogthos:
Profiling a real application's test suite (kmet, 2198 tests) against babashka found it **2.24x slower**. The gap was concentrated rather than broad: of 20.4s of total difference, 16.4s came from five namespaces, and of 117 namespaces 8 were already *faster* on jolt. Every cause turned out to be a constant factor on a linear operation — none of them an algorithm, which is why none of the existing scaling gates could see them.

## What was slow

**`(read rdr)` over a stream-backed `PushbackReader`** drained the reader one character at a time through `record-method-dispatch` — a method table found by hashing the jhost tag, then a handler by hashing the method *name*, per character — consing each onto a list to reverse and `list->string` at the end. Reading one form out of each of 332 source files cost 3909ms against babashka's 57ms, while the same parse **from a string** cost jolt 52ms against bb's 67ms. The buffering was never the difference; the dispatch was. `drain-reader` now fills in blocks.

Nothing about the buffering semantics changed: the old path drained the whole reader too, and `host-reader-read-form` already refilled the pushback reader with the unconsumed tail as a `StringReader`, so only the *first* drain was ever slow.

**`.read(char[])`** — the documented way to stream a large input without materializing it — ran the same per-character loop plus a `ja-set!` per character (a checked record accessor, a bounds test re-reading the backing length, and a cond over the backing type). It was *slower than slurping the whole file*, so the memory-bounded read cost time instead of saving it. That inversion is gone.

**`.toCharArray`** built a cons per character before the array existed, and **`(String. char[])`** did the mirror image — a chunked decode loop pays both, once per chunk.

**`clojure.string/trim`** was `(trimr (triml s))`, and because `triml` took `(substring s i len)` even when `i` was 0, trimming a string that needed no trimming allocated it **twice**.

**A great many regex calls are not regex calls.** `#"\n"` is the commonest separator in line-oriented code, and running an irregex search per line to find one character cost ~10x; `#"abc"` as a replace pattern cost ~10x against the literal arm of the very same function. A pattern that matches exactly one string is now recognized and routed to a non-allocating index scan.

The recognizer declines on any metacharacter, class, quantifier, group, anchor or `\d`-style class, and `replace` declines further on a `$`-group reference, a backslash escape, or a function replacement — anything that could mean more than itself keeps the engine. Declining is always safe; only accepting wrongly would be a bug. `split-lines` is `#"\r?\n"`, the one line-shaped pattern that is *not* a literal, so it gets its own scan.

## Numbers

Against babashka 1.12, x86_64:

| operation | before | after | babashka |
| --- | --- | --- | --- |
| `read` a form off a stream (332 files) | 3909 ms | 93 ms | 57 ms |
| `str/split #"\n"` (20k lines) | 1086 ms | 104 ms | 106 ms |
| `str/split-lines` | 1200 ms | 56 ms | 234 ms |
| `str/replace` with `#"abc"` | 1661 ms | 183 ms | 174 ms |
| `str/trim` x100k | 185 ms | 6 ms | 12 ms |
| `(String. char[])`, 8.7 MB | 457 ms | 184 ms | 22 ms |
| `.read(char[])`, 8.7 MB x20 | 5669 ms | 1509 ms | 166 ms |

End to end on the suite that motivated it, with the **same three pre-existing failures before and after**:

| | bb | before | after |
| --- | --- | --- | --- |
| test time | 16.0 s | 35.9 s | **28.6 s** (2.24x → 1.79x) |
| user CPU | 11.2 s | 29.6 s | **17.6 s** (2.64x → 1.57x) |

CPU is the honest figure — wall time includes subprocess and network waits neither host controls. The namespaces still above 2x are network/subprocess bound and swing 3.7–6.7s run to run on the *same* binary; that was measured, not assumed.

## Correctness fix found on the way

`(.read rdr cbuf)` — the one-argument `Reader.read(char[])` overload — raised `caddr: incorrect list structure` on every reader in `host-static-classes.ss`. Only the three-argument `read(cbuf, off, len)` form was ever destructured, so the shorter spelling, which means "fill the whole array", died in argument handling. `char-reader` already defaulted correctly; `string-reader` and `pushback-reader` now share that defaulting.

## Benches and gates

`bench/host_io.clj` (`host-io:200`) and `bench/string_scan.clj` (`string-scan:40`), both registered in `bench/run.sh`. Each carries **control arms** for the shapes that must not change — patterns that genuinely need the engine, and the same parse with the characters already in hand. Those are what a too-eager literal recognizer would break, so they are measured rather than assumed.

`test/fastpath_ratio_test.clj`, wired as `fastpathratio` in `CI-GATES`. Deliberately not a scaling gate: every regression here is linear with a terrible constant, which a 1x-vs-4x ratio cannot see. Each arm is a ratio against a reference measured in the same process on the same machine — self-calibrating, no absolute budget, no baseline file, no babashka in CI. Sampling follows `io_scaling_test.clj`: best-of-N minimum, a ceiling, a clear-regression threshold, and one re-measure in the band between.

**Every arm was negative-tested** by reintroducing the regression it guards, because a gate that cannot fail is worthless. Disabling the literal recognizer drives `literal-split` to 11.85 and `literal-replace` to 8.46 against a ceiling of 3.0, and the gate exits 1.

## Two things worth a reviewer's attention

**I shipped a regression mid-way and the microbenches missed it.** Sizing the drain scratch to the 64KB block made *small* reads worse than the per-character loop they replaced — 64KB allocated to hold 200 bytes, once per file. Every bench arm used a large payload, so all of them looked perfect; only re-running real namespaces caught it (900ms → 1.8s). The scratch now grows from 4KB, and both the bench and the gate gained small-file arms.

**The first version of that gate arm didn't work.** It compared `(slurp path)` over many small files against one big file — but slurping a *path* takes the bytevector route and never reaches the drain, so reintroducing the bug moved the ratio not at all (3.42 → 3.42). It now holds everything but the drain constant: the same files, the same opens, one arm through an `InputStreamReader` and the reference by path. 1.34 healthy vs 5.27 regressed, ceiling 4.0.

## Validation

- `make ci` green, 114 targets (run twice)
- 113 cross-host semantics checks across three suites — reader, `read(char[])`, and string — produce **identical results on jolt and babashka**, including `$1` group references, function replacements, CRLF and bare-CR handling, UTF-8, split limits and empty-field semantics, and the `String.`/`toCharArray` round trip
- `make attributioncheck` clean

## Known remaining

~~`String.`-from-`char[]` and `.read(char[])` keep a residual ~9x … it belongs in its own change rather than bolted onto this one.~~ **Superseded by `26535411` on this branch**, which does take the string backing: a char array is a Chez string, so `(String. ca)` is a copy of it and a reader fills one in place. The two arms that cannot hold a character — a lone surrogate, or a store of a value no char[] could hold on the JVM — fall back to a boxed vector (`ja-promote-chars!`) and behave exactly as before. The CHANGELOG carries the entry; this section was written before the commit.

`re-seq` is unchanged at ~5.4x; it genuinely needs the engine and has no literal shortcut to take.


yogthos:
Per-namespace results on the profiled suite, so the scope of this PR is clear about what it does *not* fix. Two of the five namespaces from the original profile are essentially untouched.

| namespace | bb | before | after | before | after |
| --- | --- | --- | --- | --- | --- |
| `kmet.test-http-boundary` | 295 ms | 8500 ms | **872 ms** | 28.8x | **3.0x** |
| `kmet.ai.test-oauth` | 1000 ms | 4400 ms | **2700 ms** | 4.4x | 2.7x |
| `kmet.app.test-session` | 1200 ms | 4100 ms | 4000 ms | 3.4x | 3.3x |
| `kmet.libs.test-http` | 2500 ms | 3600 ms | 5300 ms | 1.4x | 2.1x |
| `kmet.tui.components.test-markdown` | 457 ms | 1300 ms | 1400 ms | 2.8x | 3.1x |

`libs.test-http` is subprocess-bound and not a real regression — it swings 3.7–6.7s run to run on the *same* binary (measured, both directions). The other two are compute-bound and genuinely unmoved:

**`test-session`** is blocked on the residual this PR documents but does not close. `session.clj` does `(String. cbuf 0 n)` per chunk at three sites, each fed by `.read(char[])` — the two operations still at 8.8x and 9.1x. `test-session-get-tree-large-session` alone is 1904 ms against bb's 518 ms. This is the case for doing the char-array string-backing change: it is not a micro-optimization, it is what closes the last compute-bound namespace in the profile.

**`test-markdown`** is the regex engine on genuinely non-literal patterns, which the literal recognizer correctly declines. `md/parse` has 23 regex sites — anchors, classes, quantifiers, and a backreference. Worth recording that the gap is wider than the `re-seq` figure in the description suggests:

| operation (332 files) | bb | jolt | ratio |
| --- | --- | --- | --- |
| `re-find #"\[\s*\"curl\""` | 62 ms | 276 ms | 4.5x |
| `re-find #"\"curl\s"` | 10 ms | 260 ms | **26x** |
| `re-seq` over 8.7 MB | 18 ms | 98 ms | 5.4x |

The 26x case is the interesting one: `#"\"curl\s"` is a literal followed by a single `\s`. It declines the recognizer correctly — `\s` matches several different characters — and then costs 26x to find a five-character needle. A literal prefix with a small variable tail is the shape most real patterns have, and looks like the right thing to attack next.

None of this changes what is in this PR; it is the map of what is left, tracked as follow-up work.

yogthos:
Reviewed and merged `main` in — this branch and #966 both rewrote `io-streams.ss`, so the two conflicts were real ones rather than adjacency:

- **`read(byte[] …)`**: this branch reads straight into the byte array's own bytevector backing, #966 added the pushback rule (buffered bytes first, then ONE source read when they fall short of `len`). Both arms now go through `in-stream-read-some!`, which grew a start offset so the in-place arm can name the caller's `off` instead of a scratch buffer's zero.
- **`in-stream-source-port`**: kept this branch's reused growing buffer, and kept #966's `stream-read-into!` beside it — the pushback port's refill wants exactly what its own buffer asks for rather than a 64 KB read-ahead — plus `user-in-stream-bytes` / `user-out-stream-write!`.

After the merge, all nine coercion paths over a `PushbackInputStream` (`BufferedInputStream`, `io/reader`, `InputStreamReader`, `slurp`, `io/copy`, `readAllBytes`, `io/input-stream`, `line-seq`, a `proxy` source) still agree with reference Clojure 1.12.5.

Two things found in review, fixed in `1d38173`:

- `clojure.core/str-split-lines` was `def-var!`'d but answers a Scheme list, so `(clojure.core/str-split-lines "a\nb")` rendered as `#object[:object]`. Nothing called it — `str-split` is its one entry point — so the export comes out rather than the return being wrapped.
- `na-char-of`'s comment had the JVM wrong. Reference Clojure's generic `(aset chars 0 65)` is an `IllegalArgumentException`, not the widening store, and `(char-array [65])` a `ClassCastException`. `aset-char` **is** the int→char store, and that is the rule this follows and the one the `char-array` constructor already followed; the comment says so now.

Also worth a note on the PR body: the **Known remaining** section is stale. It says a char array's string backing "belongs in its own change", and `26535411` on this branch does exactly that — the CHANGELOG covers it, only the description doesn't.

Verification beyond the branch's own gates: 4000 random (pattern, input) pairs × 16 string operations — `split` at three limits, `replace`, `replace-first`, a `$0` replacement, `re-find`, `re-seq`, `split-lines`, `indexOf`, `lastIndexOf`, `.replace`, the three trims — over an alphabet seeded with every metacharacter the recognizer declines on, produce **byte-identical output on this branch and on `main`**. That is the check that matters for the literal recognizer: accepting wrongly is the only way it can be a bug.

Gates on the merge: `unit` 1819/1819, `corpus` 5696/5706 with 0 new divergences, `documented` 57/57, `ioscaling`, `arraybacking` 71/71.

PR #995, opened 2026-09-14
Fix clojure.main/repl looping on "No matching field found: read"

### LINKED ISSUE #981: Nested REPL via clojure.main/repl blows up in Jolt
What I expected:

```clojure
> clj
Clojure 1.10.3
user=> (clojure.main/repl)
user=> *clojure-version*
{:major 1, :minor 10, :incremental 3, :qualifier nil}
user=> nil
user=> *clojure-version*
{:major 1, :minor 10, :incremental 3, :qualifier nil}
user=> 
```

What happens with Jolt:

```clojure
> jolt
;; jolt v0.8.7-11-gbe356e59 repl — :repl/quit or ^D to exit
user=> (clojure.main/repl)
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
...
```

Until you `ctrl+c` to kill the process.

I ran into this trying to test `rephrase` 😄

yogthos:
Fixes #981.

```
user=> (clojure.main/repl)
No matching field found: read for class clojure.lang.IObj$reify__0
user=> Execution error (IllegalArgumentException) at (REPL:1).
No matching field found: read for class clojure.lang.IObj$reify__0
...
```
— forever, until the process is killed.

## Cause

`clojure.main/repl-read` is the reference's, written against
`LineNumberingPushbackReader`'s character protocol — `.read`, `.unread`,
`.readLine`, plus `renumbering-read`'s re-read through a fresh one. The reader
jolt binds to `*in*` speaks none of it: the default standard-input reader
(`clojure.core`, `50-io.clj`) is an `IReader` over the shared stdin buffer whose
smallest unit is a whole form, and so is the reader `with-in-str` binds, where
the JVM's `with-in-str` hands over a `LineNumberingPushbackReader`.

`(.read *in*)` on one raised, and since the raise consumed no input, `repl`'s
`:caught` hook reported it and the loop read the same nothing again.

## Fix

`repl-read` branches on the reader:

- a `java.io.PushbackReader` — which is what `clojure.lang.LineNumberingPushbackReader`
  is here — takes the ported path unchanged, so its line numbering and the
  `:read-source` positions built on it are untouched (the existing
  `clojure.main` repl unit row over an LNPBR still passes);
- anything else reads one form through `clojure.core/read`, which already blocks
  for as many lines as the form needs and answers the `:eof` value at end of
  input, so the loop ends on EOF instead of spinning.

After:

```
user=> (clojure.main/repl)
user=> (+ 1 2)
3
user=> *clojure-version*
{:major 1, :minor 11, :incremental 0, :qualifier nil}
user=> (/ 1 0)
Execution error (ArithmeticException) at (REPL:1).
Divide by zero
user=> :still-alive
:still-alive
user=> ^D
nil
user=>
```

## Divergence recorded

The one thing a form-at-a-time reader cannot offer is `request-prompt` for a
line with nothing but whitespace left — it reads on to the next form rather than
consuming past the line end to find out. Recorded in
`test/conformance/known-divergences.edn` with a machine-checked `:check`.
Nothing is lost at a REPL: `repl`'s `need-prompt` already defaults to
`(constantly true)` for an `*in*` that is not a `LineNumberingPushbackReader`,
so a fresh prompt is printed before every read regardless.

## Tests

Six new corpus rows under `clojure.main / repl over *in*` and
`clojure.main / repl-read`: evaluating every form, ending on EOF rather than
repeating, a form split across lines, per-form evaluation, `repl-read` returning
a form, and `repl-read` returning request-exit at end of input.

Gates run: `corpus` (5702/5712, 0 NEW divergences), `unit`, `documented`,
`certify` (5940 rows against JVM Clojure 1.12.5 — 0 NEW, 0 stale),
`gambitseedcheck`, `smoke`, `seeddefs`, `errorreport`. Both seeds re-minted
(`make remint` + `make gambitseed`).


yogthos:
Superseded by #984, which fixes #981 completely rather than around the edge.

#984 teaches `IReader` the three character ops `repl-read` actually wants — `-read-char`, `-unread-char`, `-at-line-start?` — all off the same cursor the line and form ops already use, and makes `clojure.main`'s `skip-whitespace` / `skip-if-eol` / `renumbering-read` / `:need-prompt` polymorphic over both reader kinds. A `java.io.PushbackReader` still takes the ported path, so the guard this PR added is subsumed.

The practical difference is the divergence this PR recorded: `repl-read` answering `request-prompt` for a line with only whitespace left. #984 answers it, so the entry in `known-divergences.edn` is not needed, and the nested REPL is now byte-identical to reference Clojure — verified against `clojure -M` on the same input.

PR #1008, opened 2026-09-15
A source-formatter bench row, and the five costs it found

yogthos:
`standard-clojure-style` is the one workload jolt has with a **non-JVM** reference: it is a Clojure port of a formatter whose upstream implementation is JavaScript, so the identical work can be timed on V8 too. Its `PERFORMANCE.md` reports ~316x V8 before the port's own optimization pass and ~6x after — but those numbers are from a 20-core machine and do not reproduce elsewhere, so the starting point here was to measure all three implementations on one box.

Same 57 KB payload, warm, one x86_64 Linux machine (4-core i5-4278U):

| | before | after | vs node |
|---|---:|---:|---:|
| node / V8 | 14.4 ms | 14.4 ms | 1x |
| JVM Clojure | 158.6 ms | 158.6 ms | 11x |
| **jolt** | **372 ms** | **286.5 ms** | **20x** |

The real `standard-clj check src/ test/ test_cases/` went **1.9s -> 1.43s** wall.

## The bench row

`bench/cst_format.clj` keeps both phases as a real formatter spells them — a 10-key map allocated per token, combinator parsers reached through `(:parse p)`, a `^`-anchored `re-find` over a window cut with `subs`, then every node wrapped in an atom, `swap! node assoc` adding ~8 keys as the walk learns each node's column, and output built by `(swap! out-txt str …)` per line. `collections` and `literals` build maps but never at one-per-input-token rates, and nothing in the suite held an atom per element of a large collection.

The row prints its own parse/format split above `mean:`, and `verify` pins both phases so neither can be optimized away or silently break. `ci/bench-gate.sh` reads the list from `run.sh`, so the release gate picks it up with no second copy.

Two rows join `test/complexity_test.clj` for the shapes the row depends on: a fixed-span `subs` and an anchored `re-find` must cost the **window**, not the file. Both are O(window) today; the gate is there so a future string representation or regex entry point cannot quietly make them O(input) while every value test still passes.

## The five costs

Each is a common case paying for a general one, and none was reachable from a value test.

- **`subs` copies the span instead of walking it.** Chez's `substring` moves one character at a time; `make-string` + `string-copy!` is ~3.7x faster for the spans a parser or reader cuts (2048 chars: 4.27us -> 1.17us, of which 1.29us is the allocation, so the copy is free and the loop is the cost). Routed from `subs`, `jolt-string-method`'s substring arm, `subSequence`, and the proven-`^String` `.substring` emit.
- **`jolt-subs` is a `case-lambda`.** The rest-arg list was most of what a one-character `(subs txt i (inc i))` cost.
- **`deref` answers an atom first.** The chain built `jolt-deref` in four `set!` layers (futures -> vars -> volatiles -> atoms), so an atom walked all four and allocated a rest list on the way in — ~60ns to read one record field. `deref` also joins `op-registry`, so `@x` lowers to a direct call; only the 1-arity form, the timed arity stays on the var.
- **`swap!` gets fixed 2/3/4-arity clauses.** No rest list, no `apply`. The retry loop is spelled out per clause rather than shared through a thunk — a thunk trades the list for a closure and gives the fixed arities back what they came for (measured both ways).
- **`amap-index` gets a `char?` arm, and `case` compares char constants by identity.** A set of characters is how Clojure spells a character class and a scanner probes one per position; both were on generic-equality paths. The `case` one is easiest to see inverted: replacing a linear `(.indexOf " ,\n\r\t\f" c)` with `(case c …)` over those same six characters used to make a scanner **slower**, 174.6 -> 200.6 ms. It is now 164.7, so the idiomatic spelling is also the fast one.

Note `bench/char-scan` does **not** move — its `case` dispatches on small integer states, not characters.

## Results

A/B, min of three timed runs per side, baseline built from the parent commit:

```
bench              baseline  candidate    ratio
cst-format            372.9      287.8    0.77x
host-io              3233.0     2480.0    0.77x
stm                   273.8      254.6    0.93x
printing             1857.5     1783.0    0.96x
collections/seqs/transients/keyed-lookup/hash-eq/
string-ops/string-scan/char-scan/literals          flat
```

Gates: unit 1878/1878, corpus (0 new divergence), certify, documented, complexity incl. the two new rows, fastpathratio, all twelve scaling gates, threadsafety, stateimage, fibers/gosm/flow/asynctimer, gambitcheck/gambitboot/gambitseedcheck/gambitgencheck, manifest, and the build and packaging gates. sci's 12 known failures are unchanged from the parent. Both seeds re-minted.

## What this does not do

The target was "comparable to JS, at worst under an order of magnitude", and this does not get there — jolt is still ~20x V8 on the row. The reason is worth recording: I wrote the same work the way the JS writes it (a `deftype` with mutable fields, a `StringBuilder`, a hand scanner, no regex window) and it lands at **165 ms, still ~11x V8**. Since the JVM is also ~11x V8 on the idiomatic version, the ordering is that the idiom costs ~11x before jolt is involved, jolt costs ~1.8x the JVM on top, and rewriting the port to be JS-shaped buys back roughly that 1.8x and no more. Closing further means jolt-side work on deftype field access, `nth`, and per-character interop — the `cst-format` row now exists to measure it.


yogthos:
Reviewed and re-measured on an M1 Pro (macOS), A/B/A against main, min of 5 runs per side, `--direct-link --opt` builds from each tree:

```
bench            A1     B     A2    B/A
cst-format    142.9  120.9  141.9  0.85x
stm           122.5  119.3  123.8  0.97x
keyed-lookup     94     92     94  0.98x
string-ops      283    283    281  flat
char-scan       113    113    112  flat
collections    16.8   17.0   16.9  flat
host-io        1029   1022   1065  flat   (the 0.77x in the description does not reproduce here)
string-scan    5226   5178      -  flat
printing      792.6  789.6  814.2  flat
seqs          348.8  352.6  354.1  flat
transducers   125.3  126.5  130.9  flat
```

One real bug: `jolt-substr` called `string-copy!` directly, and converters.ss is in the gambit boot. Chez's order is `(from from-start to to-start count)`, R7RS's is `(to at from start end)` — same arity, so it compiled on gsi and `(subs "hello world" 6)` returned five NULs and overwrote the source. Every Chez gate was green; the portcheck blocklist header even called `string-copy!` "standard R6RS" (R6RS has no `string-copy!`). 30aca5fe adds `sa-string-copy-range!` to the adapter contract, routes every host `string-copy!` through it, blocklists the raw name, and adds a gambitcheck adapter row plus a gambitkernel `subs` row that fails against 2d43f6dd.

Also in that commit: `vswap!` gets the same fixed arities as `swap!` and `deref`'s fast arm answers a volatile (flat on `transducers`, so rest-list only); five gambit mirrors that `make gambiteval` (not in ci, dead since 2026-09-02) turned up — the rest of that gate is jolt-cw2p; and the bench header's "~40x V8" brought in line with the README's ~20x.

Gates on the amended branch: `make test` (114 ci + 3 test targets), `make libconformance` 47/47 ok 0 regressed, gambitcheck/gambitkernel/gambitboot/mirrordrift/deadhost/portcheck/adaptercheck.

PR #1032, opened 2026-09-16
Drop four process-wide locks from per-call paths

yogthos:
Found re-running standard-clojure-style-jolt's PERFORMANCE.md on main: eight fibers on eight carriers gave 1.7x where eight separate processes doing the same work gave 6.5x. A per-op microbench (`future` x1 vs x8, each thread on its own values, GC rendezvous excluded with a 1 GB nursery) and macOS `sample` put it on Scheme-level mutexes:

| op, per-thread slowdown at 8 threads | before | after |
|---|---|---|
| `assoc` on a 6-key map | 32x | 3.8x |
| `swap!` assoc on an own atom | 21x | 2.8x |
| `str` of a Long / `Long/toString` | 17x / 20x | 1.4x / 1.2x |
| `re-find` with a `#"…"` literal | 24x | 2.0x |
| `hash` of a fresh string | 12x | 2.1x |
| `keyword` of `(str "k" n)` | 18x | 1.1x |

Arithmetic, cons, `subs`, `.charAt`, `get`, `=` scaled fine throughout.

## Causes and fixes

- **Metadata side-table read under its mutex on every `conj`/`assoc`.** `meta-carry` probed the table for the receiver's meta; the "no metadata anywhere" fast path counted insertions and never came back down — 66 after the prelude alone with the table already empty — so it was dead in every real program, and the probe was a third of `assoc` and half of `conj` single-threaded. The read is a seqlock now (`meta-gen` odd while a writer is inside, fences, retry under the mutex when the generation moved across the lookup) and the empty check reads `hashtable-size`, which Chez keeps current for a weak table as the collector drops keys. A plain unguarded read loses a carry about 1 run in 3 — `make threadsafety` scenario 16 pins that (8 threads x 40k with-meta/conj/meta, verified failing with the guard removed). Single-threaded: `assoc` 98 → 69 ns, `conj` 68 → 42, `meta` 51 → 25.
- **Every `letcc` capture registered in a weak table under a global mutex** so `escape-fn?` could recognise the wrapper. An escape is now recognised by its code object's name: `sa-procedure-code-name` (new adapter/contract entry; Chez `$code-name` of `$closure-code`, which survives no-inspector builds and is the same for every closure of one lambda), and the wrapper is bound under `jolt@escape`, a name no Clojure symbol can spell. The formatter's parser takes a letcc per `Choice` attempt; eight carriers spent more than half their samples on that lock.
- **A regex literal's cache hit took the compile lock.** `#"…"` evaluates to `(jolt-regex "src")` each time it is reached; the hit is lock-free now, double-checked under the mutex on a miss. Hoisting the literal per site (the JVM's constant Pattern) is jolt-ax00.
- **Chez's `number->string` is `(format "~d" x)`** — ~230 ns for a fixnum and behind the formatter's one lock (`format`/`printf` serialize the same way). `jolt-fixnum->string` (values.ss, a digit loop, 29 ns; checked against `number->string` on the fixnum edges and 200k randoms) now serves `str`, the printer and `Long/toString` radix 10; Gambit's `rt-core.ss` mirrors the arm. `str` of a Long 261 → 55 ns, `Long/toString` 221 → 98.
- **The per-thread string-hash cache was weak.** A string hashed once and dropped left an entry every collection traced; it is a bounded strong table now (cap 2048, clear on overflow), the shape the symbol cache already took.

`jolt-mt?` moves to values.ss: natives-meta.ss and seq.ss load before lazy-bridge.ss, which defined it, and the Gambit boot's pre-declaration for the same reason goes.

## What remains

`with-meta` itself still serializes (a write into the side table); the fix is the JVM's meta field on the collection records, which is image-format surface — jolt-xsgu. With the Scheme locks gone, an allocation-heavy parallel workload sits in Chez's own allocator: `S_get_more_room` refills a thread's allocation area one 16 KB segment at a time under `alloc_mutex`, with `S_scan_dirty` inside — jolt-3o2t.

## Gates

`make test` exit 0 (121 ci targets + selfhost), corpus 0 new divergences with two new integer-rendering rows, `make threadsafety` 3/3, gambit gates + mirrordrift + adaptercheck + portcheck. bench-gate and libconformance results follow in a comment.


yogthos:
`ci/bench-gate.sh` against a release build of main (58b174a9), min of 3 runs per side:

| bench | baseline | candidate | ratio |
|---|---|---|---|
| collections | 17.0 | 17.9 | 1.05x |
| lazy-threads | 242.4 | 243.7 | 1.01x |
| keyed-lookup | 93.0 | 93.0 | 1.00x |
| hash-eq | 461.0 | 465.0 | 1.01x |
| literals | 53.0 | 53.0 | 1.00x |
| string-build | 180.0 | 180.0 | 1.00x |
| string-ops | 279.0 | 282.0 | 1.01x |
| char-scan | 110.0 | 109.0 | 0.99x |
| printing | 798.3 | 753.4 | 0.94x |
| dispatch | 66.9 | 66.0 | 0.99x |
| executors | 1324.5 | 1320.0 | 1.00x |
| cst-format | 120.7 | 121.8 | 1.01x |

All within the gate's noise band (~1.07x). libconformance is running.

yogthos:
`make libconformance` on this branch's release binary: 50 libraries: 47 ok, 0 better, 0 regressed, 3 skipped/missing.

yogthos:
End to end on the program that surfaced this (standard-clojure-style-jolt's `standard-clj check` over its twelve files, M1 Max, release builds), with #1034 and #1035 stacked on top:

| build | reported total (median of 5) | wall clock |
|---|---:|---:|
| main | 464 ms | 0.52 s |
| main + #1032 + #1034 + #1035 | 364 ms | 0.42 s |

Single file, `parse_ns.clj` (78 KB): 209 → 182 ms; `format.clj` 141 → 122; `parser.clj` 40 → 34. The port's own suites (141/83/236) pass on the combined build. What is left in that program is its allocation profile (721 MB per format of the 78 KB file, GC 7%), not a lock.

Issue #1006, opened 2026-09-15
SCI can't implement a `copy-var*`-injected host protocol — follow-up to #1000


markokocic:
**Build:** `v0.8.1-443-gd55ec029` (main through #1003). SCI 0.13.53 (jolt-vendored, same version kmet resolves).

**Background:** #1000 closed with #1003 (`~@`-laziness + `:sigs`, recipe pinned by `sci-functional-test`). The recipe works on this build (verified: record, deftype, and extend through the multimethod mirror all return correct values). The transparent flow — `defrecord`/`extend-type` directly over a host protocol shared by reference — still fails. The #1000 commit message declines the fix (`:ns` "is not added … a jolt namespace would not satisfy SCI's HasName anyway"). This ticket is that remainder.

### Repro (self-contained, no app code)

```clojure
(require '[sci.core :as sci])
(defprotocol Shape (area [this]) (scaled [this k]))
(let [sns (sci/create-ns 'my.host)
      varmap (into {} (keep (fn [[k v]]
                              (when (and (var? v) (#{'Shape 'area 'scaled} k))
                                [k (sci/copy-var* v sns)])))
                   (ns-interns *ns*))
      ctx (sci/init {:classes {:allow :all}
                     :namespaces {'my.host varmap}})]
  (sci/eval-string* ctx
    "(defrecord Sq [s] my.host/Shape
       (area [_] (* s s)) (scaled [_ k] (->Sq (* s k))))"))
;; jolt: FAIL "Unable to resolve symbol: area" (phase analysis)
```

```clojure
(sci/eval-string* ctx
  "(extend-type String my.host/Shape
     (area [s] (count s)) (scaled [s k] s))")
;; jolt: FAIL "No method getName in sci.impl.types/HasName"
```

Parity evidence (identical injection code + identical eval string): a host protocol injected under its own ns name with all method vars present builds and round-trips on bb (`construct → satisfies? → call inside → host satisfies? → host call`, all green) and fails on jolt with the two errors above. SCI-defined protocols on jolt are fine (same-string, cross-ns via `:load-fn`, aliased calls all evaluate), so this is specifically the *injected host* shape.

### Root cause (three seams, all in vendored SCI 0.13.53)

SCI's record/type codegen hardcodes the JVM protocol shape. Jolt's shape supplies none of it:

| SCI call site | needs | bb/JVM value | jolt value (`make-protocol`, `host/chez/protocols.ss`) |
|---|---|---|---|
[...]

PR #1033, opened 2026-09-16
Complete clojure.lang.Var's instance surface, including the two members the JVM does not have

### LINKED ISSUE #1031: SCI `IVar` protocol missing `:getRawRoot` for `clojure.lang.Var`
**Environment:** jolt v0.8.8-53 (or current main), Linux/aarch64 (Termux), also reproducible on Linux/x86_64
**Area:** Vendored SCI (`sci.impl.vars`), jolt's `clojure.lang.Var` shim

## Summary

The `IVar` protocol in vendored SCI 0.13.53 (`sci.impl.vars`) declares a `:getRawRoot` method that is expected to be implemented on `clojure.lang.Var`. jolt's `Var` shim (in `host/chez/java/vars.ss` or similar) does not implement this method, causing all code that uses `sci/copy-var*` or SCI's var introspection to fail when loading namespaces that define `defrecord`/`defprotocol`/`deftype` over host protocols.

## Error

```
No implementation of method: :getRawRoot of protocol: #'sci.impl.vars/IVar
found for class: clojure.lang.Var
```

## Repro (minimal, self-contained)

```clojure
;; Start jolt repl or run with -e
(require '[sci.core :as sci])
(require '[kmet.tui.protocols :as p])  ;; any host protocol

;; Inject the protocol into a SCI namespace (same as kmet.extension does)
(def sns (sci/create-ns 'my.host))
(def varmap (into {} (keep (fn [[k v]]
                             (when (and (var? v) (#{'IComponent 'render 'handle-input 'invalidate} k))
                               [k (sci/copy-var* v sns)]))
                          (ns-interns *ns*)))
(def ctx (sci/init {:classes {:allow :all}
                    :namespaces {'my.host varmap}}))

;; This fails on jolt, works on babashka/JVM
(sci/eval-string* ctx
  "(defrecord R [] my.host/IComponent
     (render [_ w h] nil)
     (handle-input [_ e] nil)
     (invalidate [_] nil))")
;; => FAIL: No implementation of method: :getRawRoot of protocol: #'sci.impl.vars/IVar found for class: clojure.lang.Var
```

## Root Cause

In SCI 0.13.53 (`src/sci/impl/vars.cljc`), the `IVar` protocol is defined with `:getRawRoot`:

```clojure
(defprotocol IVar
  (getRawRoot [this])
  ...)
```

And `copy-var*` calls `IVar/getRawRoot` on the source var:

```clojure
(defn copy-var* [v target-ns]
  (let [root (IVar/getRawRoot v)]  ;; <-- this call fails on jolt
    ...))
```

On bb/JVM, `clojure.lang.Var` implements `IVar` via `extend-type` in SCI's initialization. On jolt, the `Var` shim (which wraps Janet's var-like objects) does not have this protocol extension.

## Impact

**Blocks all kmet shipped extensions on jolt:**
- `extensions/mcp-adapter` (defcomponent panels)
- `extensions/lsp-adapter` (defcomponent panels)
- `extensions/review` (defcomponent dialogs + extend-type IFocusable)
- `extensions/clojure` (defcomponent + Maven-chain gaps)
- `extensions/tree-sitter` (loads but only because it has no defcomponents — any added component would fail)

This is **distinct from jolt#1006** (which was about `defrecord` analysis over injected host protocols — that fix landed in #1003). This is a missing protocol implementation on jolt's `Var` shim.

## Suggested Fix

Add `IVar/getRawRoot` implementation to jolt's `clojure.lang.Var` shim (likely in `host/chez/java/vars.ss` or wherever the Var shim is defined), returning the underlying root binding — analogous to what `clojure.lang.Var.getRawRoot()` returns on the JVM (the non-thread-local root value).

## Workaround in kmet

`^:bb-only` gate on `test-shipped-extensions-load-from-src` (test/kmet/app/test_extensions.clj:1056) — extensions only tested on babashka.

## Verification

Once fixed, this should pass on jolt:
```bash
jolt test kmet.app.test-extensions/test-shipped-extensions-load-from-src
```

And the minimal repro above should evaluate without error.

yogthos:
Fixes #1031.

## What the report asked for, and what is actually wrong

The issue asks for an `IVar/getRawRoot` implementation on jolt's `Var` shim. That
part of the diagnosis does not hold up: `(.getRawRoot v)` has worked for a long
time, and `sci/copy-var*` never calls it — it reads `(meta v)` and `@v`.

The error comes from `sci.impl.vars/alter-var-root`, which calls `getRawRoot`
**through the `IVar` protocol**. Nothing extends `IVar` to `clojure.lang.Var` —
not SCI, and not Clojure. Verified against Clojure 1.12.5 with this same
vendored SCI:

```
JVM 1.12.5:  defrecord over a host protocol copied as-is
             => No implementation of method: :getRawRoot of protocol:
                #'sci.impl.vars/IVar found for class: clojure.lang.Var
```

That is the string jolt already produces, and `make scifunctional` has pinned it
since #1006. The reporter sees it work on babashka because under bb the "host"
namespace is itself SCI code, so its vars are already `sci.lang.Var` and there is
no host/SCI boundary to cross. On jolt — as on the JVM — the program is
host-native, so there is one.

## The real gap this exposes

Since SCI expects the *embedder* to supply the `IVar` extension, the thing that
matters is whether that extension can be written. On the JVM it is seven lines of
`clojure.lang.Var` methods. On jolt it did not compile, because `.toSymbol` — the
method `IVar` literally names — raised `No matching field found`.

Probing the whole surface turned up five members the JVM has and jolt did not:

| member | JVM | jolt before |
|---|---|---|
| `(.toSymbol v)` | `user/thing` | No matching field found |
| `(.-ns v)` / `(.-sym v)` | public final fields | No matching field found |
| `(.unbindRoot v)` | root → `Var$Unbound` | No matching field found |
| `(.set v x)` | writes the thread binding | No matching method |
| `(.fn v)` | `(IFn) deref()` | No matching field found |

## The change

`host/chez/records-dispatch.ss` (shared with the Gambit boot; `records-gambit.ss`
regenerated via `make gambitgen`):

- **`toSymbol`** — `Symbol.intern` of ns.name and sym.name.
- **`ns` / `sym` take the dashed spelling.** They are public final fields, so the
  JVM reads each as both `(.ns v)` and `(.-ns v)`. Only these two take the dash;
  a dashed spelling of a *method* is a field read and stays a miss, which a
  corpus row pins with `(.-getRawRoot v)`.
- **`unbindRoot`** — resets the root to the var's `Unbound` marker through the
  same `var-root-set!` / `make-jolt-var-unbound` pair `ns-unmap` already used.
- **`set`** — `jolt-set-var!`, the one write path already behind the `set!`
  special form and `var-set`: validator first, thread binding only, and
  `"Can't change/establish root binding of: x with set"` with none in place.
- **`fn`** — `deref` cast to `IFn`, so a var holding a non-fn reports the cast
  rather than handing back a value the caller then fails to invoke.

Arity is guarded on the three new method arms, so a wrong-arity call still falls
through to `dispatch-miss` and reports the JVM's own "No matching method".

## What this does and does not fix

It makes the embedder-side `IVar` extension **writable on jolt, verbatim from the
JVM**. It does **not** make a host protocol copied into SCI as-is work — nothing
could, short of SCI changing, because the failure continues on the JVM. With
`IVar` extended, both hosts now get one seam further and land on the same error:

```
jolt (this branch): Unable to resolve symbol: area
JVM 1.12.5:         Unable to resolve symbol: area
```

So for the reporter: the `^:bb-only` gate on `test-shipped-extensions-load-from-src`
should stay. The supported way to share a host protocol into SCI is the
multimethod recipe that `test/chez/sci-functional-test.clj` runs end to end
directly below the rows in question.

## Second commit: the two members that were left over

Both were filed as follow-ups when this branch opened and are now folded in, so
the surface is finished in one place.

### `.getThreadBinding` (`jolt-3ss`)

The only member that needed a value type, which is why it was scoped out of the
first pass. It answers the box the innermost thread binding lives in, or `nil`
with none in place.

The box is opaque, deliberately: `Var$TBox`'s `thread` and `val` fields are
package-private on the JVM, so `(.val b)` and `(.-val b)` raise
`No matching field found` there and now here too. That leaves nil-vs-not,
`(class b)`, and identity — and identity is the part that has to be right:

| | JVM 1.12.5 | jolt, this branch |
|---|---|---|
| two reads in one `binding` | `identical?` | `identical?` |
| a read from an inner frame | not `identical?` | not `identical?` |
| after `(set! *v* x)` | same box | same box |
| `(str b)` | `clojure.lang.Var$TBox@<hex>` | `clojure.lang.Var$TBox@<hex>` |

On the JVM the frame *is* a map of var to box. jolt's frame entry is the mutable
`(cell . value)` pair `dyn-walk-frames` already finds, and that pair has exactly
that lifetime — one per frame entry, alive as long as the frame, mutated in place
by `jolt-set-var!`. So the box is interned **off the pair** rather than built per
call; a fresh box per call would fail the first three rows above.

Interning is lazy and behind a mutex, and nothing in jolt or `clojure.core` calls
`getThreadBinding` (`IVar` does not name it either), so a binding nobody asks
about never allocates a box and `binding` itself is untouched. The table is
global-behind-a-mutex rather than per-thread because the pairs it keys on are
already thread-local — `dyn-binding-stack` is a thread parameter, and binding
conveyance pushes a frame built from a *copied* map, not from these pairs — while
the table is the one piece that would otherwise be unsynchronized cross-thread
mutation. Weak keys, so a popped frame's boxes go with it.

`host/chez/vars.ss` owns the record type and its `class` / `str` / `pr-str` arms,
beside `Var$Unbound`'s; `host/chez/dyn-binding.ss` owns the interning.

### `.getName` / `.name` now raise (`jolt-ggd`)

These were jolt-only aliases — `getName` for `toSymbol`, `name` for `sym` — and
Clojure 1.12.5 answers neither, because `Var` is not `Named` and has no
`getName`:

```
JVM 1.12.5:  (.getName #'plain) => IllegalArgumentException
                                   No matching field found: getName for class clojure.lang.Var
             (.name    #'plain) => IllegalArgumentException
                                   No matching field found: name for class clojure.lang.Var
```

Answering where the reference refuses is the worse half of an interop gap: code
written against jolt reads them and then breaks on the JVM, with nothing here to
have warned it. Both spellings the JVM *does* have (`toSymbol`, `.sym`) are
answered, so nothing is lost. `(name v)` and `(symbol v)` are untouched.

**Caller survey**, since removing them is a compatibility break. Rather than
grepping for a dynamic member call, each alias was made to log its receiver and
every gate run over the instrumented build:

| surveyed | hits |
|---|---|
| `corpus` (5807 rows), `unit`, `values`, `smoke`, `scifunctional` | 0 |
| `libconformance` — 45 runnable libraries incl. spec.alpha, tools.reader, core.logic, test.check, instaparse, malli, reitit | 0 |
| `selfhost` (mint + fixpoint) | 0 |
| `grenadine` | 0 |

Nothing in jolt, `clojure.core` or the vendored libraries reads either.

## Tests

- **15 corpus rows** total under `host-interop / Var surface` (7 from the first
  commit, 8 from the second), every one certified against reference Clojure
  1.12.5 — `make certify`: 6043 rows, **0 NEW, 0 stale**.
- Each of the 8 new rows was **proven to run** by planting a sentinel expected
  value and watching the gate fail on it, rather than trusting a green run.
- **`make scifunctional`** — the #1006 rows are unchanged, plus two checks from
  the first commit: the `IVar` extension compiles and answers through `Var`'s own
  methods, and the copied-as-is path dies at the JVM's next seam. Both also
  verified to fail loudly when the expected value is altered.
- The same 20-probe oracle script run on jolt and on Clojure 1.12.5 now prints
  **byte-identical output**, including every raise.

Gates run locally, all green: `certify`, `corpus`, `scifunctional`, `unit`,
`values`, `smoke`, `selfhost`, `libconformance`, `grenadine`, `gambitgencheck`,
`gambitseedcheck`.

Gambit is not installed on this machine, so `gambiteval` is skipped for CI. The
generated `records-gambit.ss` is current, and every helper the new code uses
exists on that host: `dyn-find-binding` / `jolt-set-var!` / `var-cell-deref` via
`boot.ss`'s `##include` of `dyn-binding.ss`, `jolt-identity-hasheq` in
`hasheq.ss`, and `make-weak-eq-hashtable` / `make-mutex` / `jolt-with-mutex` /
`hashtable-ref` / `hashtable-set!` in `prelude-shims.ss`.

## Follow-ups

None. `jolt-3ss` and `jolt-ggd` are closed by this branch, and
`clojure.lang.Var`'s instance surface now matches Clojure 1.12.5 member for
member — including the two members it does not have.


yogthos:
Review. The surface itself is right; probing the members it adds against Clojure 1.12.5 with the same program on both runtimes turned up three divergences, fixed in the follow-up commit.

**`(.fn v)` refused every IFn that is not a fn.** The arm tested `fn?`, so a var holding a keyword, symbol, map, set or vector threw the cast; `Var.fn` is `(IFn) deref()` and all of those are `IFn`. It asks `clojure.core/ifn?` now, and the cast it throws names the class ("class java.lang.Long cannot be cast to class clojure.lang.IFn"), as the JVM's does.

**`set!` from a non-binding thread wrote into the parent's binding.** The description says conveyance "pushes a frame built from a copied map, not from these pairs"; it does not — a future, an agent action and a go block restore the parent's stack as it is, pairs included (which is also why `getThreadBinding` from a future answers the parent's box, `identical?`, exactly as `Frame.clone` sharing the JVM's TBoxes does). So `(binding [*v* 2] @(future (set! *v* 3)) *v*)` read 3 here, an unsynchronized write on a pair the parent thread is reading, where the JVM refuses "Can't set!: *v* from non-binding thread". A frame now records its owner (the fiber running there, else the thread — `Var$TBox.thread`), and `set!` / `var-set` / `.set` refuse a frame the current thread did not push. `binding`, `bound-fn` and `with-bindings` in the child push their own and may `set!` that, as before. Bench (`make dynbench`, base / branch / base): bound read at depth 512 1500 / 1546 / 1506 ns, absent read flat at 6.4, one-var push 90 / 94 / 93 — one `cdr` per frame on the walk.

**`Thread.start` and executor tasks conveyed bindings.** Both restored the caller's stack into the worker; the JVM conveys only through `binding-conveyor-fn` (`future`, `send`, `pmap`, `bound-fn`, core.async), and a raw `Thread.` or an `ExecutorService` task starts with none. They start with an empty stack now (Chez seeds a forked thread's parameter from the forker, so it has to be explicit). The smoke row `conveyed to a raw thread` pinned the old behaviour — it wanted 8 where the reference reads the root — and is corrected, with a `bound-fn` row for the path that does carry it.

Also fixed on the way: `make dynbench` died at its first push ("Can't dynamically bind non-dynamic var: bench.dyn/*p1*") because the harness stamped `:dynamic` metadata without the flag the runtime binds on; it goes through `set-var-meta!` now. The TBox comment's claim about thread-local pairs is corrected. Gambit gets a `jolt-current-fiber` stub (`#f`), since its boot includes dyn-binding.ss but not fibers.ss.

Twelve corpus rows certified against 1.12.5 (`make certify`: 6055 rows, 0 new, 0 stale). Gates run here: certify, corpus, unit, values, smoke, scifunctional, threadsafety, gambitgencheck.

Filed on the way, not fixed here: `ScheduledExecutorService.schedule` / `scheduleAtFixedRate` / `scheduleWithFixedDelay` are missing (jolt-hgjo) — `newSingleThreadScheduledExecutor` answers a plain pool.

PR #1036, opened 2026-09-16
Collection metadata is a field of the collection record

yogthos:
A vector, map, set, list cell, `()` and lazy-seq node now carry their metadata in a slot of their own record — the reference's `Obj._meta` (`PersistentVector`, `PersistentHashMap`, `PersistentHashSet`, `PersistentList`, `Cons`, `LazySeq`, `EmptyList` all extend `Obj`). Until now every collection's meta lived in one identity-keyed side table that every `conj`/`assoc`/`into` probed (lock-free since #1032, but still a shared table) and every `with-meta` wrote under one mutex. `meta` is a field read, `with-meta` one allocation, a carry copies the slot. The side table stays for what still needs it: records, reifies, fns, sorted collections.

The slot is mutable so the copy `with-meta`/`meta-carry` build and the object the image walker rebuilds can be filled in before anyone else can see them; `coll-meta-set!` is the only writer and every caller holds a fresh instance. `with-meta` still always returns a copy.

**Image format 8.** The six layouts are image surface (chez-pvec-v4, chez-pmap-v6, chez-pset-v3, chez-cseq-v7, jolt-lazyseq-v3, empty-list-v3). Images of format 7 and older still read: the slotless records re-mint through legacy arms, and the sidecar — which in an old image holds every collection's meta, as legacy pmap records itself — now rides through the restore walk together with the body (one memo, so a pair keys the object standing in the restored graph) and re-attaches afterwards. A permanent format-7 fixture written by the pre-change build (`image-v0.8.8-coll-meta.image`: every kind with and without meta, nested, a >32 vector, a hash-mode map, meta'd keys and set elements, a record) proves it. The three container walkers walk the slot like any field, so a rebuilt map takes the walked meta.

**Semantics.** The identity keying could not express three things the JVM does, and the corpus now pins them (11 JVM-certified rows):
- `pop` of a list answers the tail node with the tail's own meta (`PersistentList.pop` returns `_rest` as is, `EMPTY.withMeta(meta())` only when the rest is empty). jolt stamped the receiver's meta onto the shared tail cell, so `(rest l)` changed too.
- `conj`/`into` onto a Cons or a lazy seq carry none (`ASeq.cons` is `new Cons(o, this)`); a list and `()` still carry.
- `vec` of a vector drops its meta (`(with-meta coll nil)` in `clojure.core/vec`), and `with-meta` on a lazy seq forces it and shares the forced seq (`new LazySeq(meta, seq())`).

**Measured** (M1 Max, quiet, dev images, A/B/A vs main at 705ac0b9; final pass):

| op | main | this PR |
|---|---|---|
| `with-meta` on a vector | 214 ns | 36 ns |
| `(meta (with-meta v m))` | 469 | 53 |
| `assoc` on a meta'd map (carry) | 333 | 99 |
| `conj` on a meta'd vector / list (carry) | 396 / 435 | 71 / 61 |
| `vary-meta` | 624 | 315 |
| `meta` on a map | 26 | 18 |
| `assoc` 6-key map, `conj` vector, `conj` list (no meta) | 91 / 55 / 34 | 83 / 49 / 28 |

8 threads on their own values, per-thread slowdown vs 1 thread: `with-meta` 34x → 2.0x, `with-meta`+carry 36x → 1.9x, `assoc` on an own meta'd map 31x → 2.0x, and `assoc` with no meta at all 12.4x → 4.5x (main probed the shared table on every assoc once any meta'd value existed in the process).

The no-meta carry had to be one call: a top-level procedure call is 4–6 ns on this path and the first cut made two or three (`conj` on a list read 32 → 42 ns); both carry helpers now test the kind with the record predicate and read the slot inline.

Gates: `make test` (all 121 targets), certify 0 new / 0 stale, libconformance 47 ok / 0 regressed / 3 standing skips, `ci/bench-gate.sh` main-release vs branch-release 41 rows within noise (max 1.04x). standard-clojure-style-jolt on a build with this jolt: output byte-identical on all 12 inputs, its suites 141/83/236, and its twelve-file time unchanged (it carries no collection metadata).


yogthos:
Review. The slot, the carries, the image format bump and the legacy path all hold up; the format-7 fixture is the right proof. Two divergences in the code this PR now owns, fixed in the follow-up commit, both certified against Clojure 1.12.5:

**`with-meta` on a seq cell ran its unforced tail twice.** `coll-with-meta`'s cseq arm copied the tail word as it stood (as `meta-copy` did before it), so a tail still pending — the thunk a lazy `map`/`filter`/`rest` leaves in the cell — was run once by the original and again by the copy: `(let [s (seq (map f (list 1 2 3))) t (with-meta s m)] (doall s) (doall t))` called `f` five times, where `Cons.withMeta` shares one `_more` and the JVM calls it three. Chunked cells were fine (the chunk is shared), which is why the corpus did not see it. The copy's tail is now a `rest` descriptor over the original cell (`lz-rest`, the producer `jolt-rest` already registers), so the producer runs once whichever cell is walked first, and the copy still dumps to an image — a state-image row round-trips one. The mirror flag follows the word.

**`vec` of a map entry handed the entry back.** The arm this PR rewrote returned a pvec as itself, and an entry is a pvec of kind `#t`, so `(class (vec (first {:a 1})))` stayed `MapEntry`; `MapEntry` is not `IObj`, so the reference's `vec` builds a `PersistentVector` of the two slots. One row.

Three comments in state-image.ss still said the sidecar entry "lands via image-meta-copy!" — since the restore walk now carries the sidecar with the body, it lands through `image-reattach-meta!` after the walk, and the legacy arms carry nothing; corrected.

Gates run here on the branch: certify (3 new rows, 0 new divergences), corpus, stateimage, unit, values, smoke, threadsafety.

PR #1037, opened 2026-09-16
Open-code record predicates and accessors

yogthos:
Record predicates, accessors, mutators and constructors are open-coded

A nongenerative record type's rtd is a compile-time constant in Chez, and cp0 open-codes `(record? v 'rtd)`, `(record-accessor 'rtd i)` and `(record-constructor rcd)` on a constant rtd. It cannot see through the top-level variable `define-record-type` binds each generated name to, though: the runtime is a sequence of top-level forms, so every `(pvec? x)` in a later form was a call through the symbol's value cell, and a kind-dispatch chain paid one call per arm (jolt-zvb7: a failing `pvec?` arm ahead of the matching one cost a list conj 6 ns).

The Chez adapter now wraps `define-record-type`: after the R6RS form has bound the names as variables, each becomes syntax that expands to the open-coded form in operator position and to a `%rec/` alias of the procedure as a value. Wrong-arity calls fall through to the procedure (same runtime error as before), `set!` of a generated name is a syntax error, a generative type or a protocol/child constructor keeps the plain procedure, and the variable binding stays so a file compiled before the type's definition still resolves. cp0 also merges a predicate's test into the accessor that follows it, so `(if (pvec? x) (pvec-cnt x) …)` is one type test.

Because a reference compiled before the definition stays a call, the seven collection layouts (pvec, pmap, pset, cseq, empty-list-t, jolt-lazyseq, jrec) move to values.ss, the first shared file, ahead of `jolt=`, `jolt-hash`, hasheq.ss and the collection dispatchers that test them; the constructors that fill defaults stay with their operations. The Gambit boot includes values.ss first as well, and records-gambit.ss is regenerated. Image format is unchanged: same tags, same fields.

`make recordinline` (test/chez/record-inline-test.ss) pins three things: the expansion shape (no call to a generated name remains, a predicate then an accessor is one type test), the load-order invariant (values.ss, hasheq.ss, collections.ss, seq.ss and natives-meta.ss name no record op a later-loaded file defines, except a per-file list of cold arms checked for staleness), and the cost as a ratio measured in one process: a six-arm kind chain and a predicate-guarded accessor, open-coded against the same chain through the procedures the names denote as values (what every generated name was before). 0.45 with the wrapper, 1.00 on the tree without it; ceiling 0.75.

Three bench rows for what this round changed, each portable and each moved by it: `coll-dispatch` (kind dispatch on small collections), `metadata` (with-meta/meta/vary-meta, carries, a positioned form tree rebuilt with meta kept) and `parallel-colls` (eight threads on their own values; prints the per-thread slowdown). README.tmpl says which rows cover the round.

Measured, release binaries, this checkout's base (perf/meta-field tip) vs this branch, alternating runs on one machine:

| op (release binaries, ns/op) | base | this branch |
|---|---:|---:|
| `get` hit, 4-key map | 11.8 | 7.0 |
| `assoc` existing key, 4-key map | 67 | 52 |
| `assoc`, 40-key map | 256 | 220 |
| keyword invoke on a map / on a record | 26.6 / 55 | 21.2 / 46 |
| `get` on a record | 40 | 31 |
| `conj` vector / list / set | 42 / 25 / 73 | 30 / 17 / 54 |
| `first` list / vector | 4.2 / 11.2 | 2.2 / 5.4 |
| `rest` / `next` of a list | 17.7 / 18.7 | 6.6 / 8.8 |
| `nth` vector / `seq` vector | 12.1 / 24 | 6.8 / 15.6 |
| `count` list / map / vector | 57 / 5.8 / 4.9 | 24 / 2.6 / 3.2 |
| `=` two vectors / two keywords | 55 / 9.4 | 37 / 5.8 |
| `hash` vector / keyword | 11 / 6.1 | 7.4 / 4.4 |
| a record predicate or accessor, hit or miss | 2.1 (1.3 net of the loop) | 0.8 (the loop) |
| `meta-carry` on a list cell (no meta) | 7.7 | 3.5 |

standard-clojure-style built with each binary (`--opt --closed-world --direct-link`), same source: the twelve-file `check` 383 → 337 ms (median of 5, alternating), `parse_ns.clj` alone 190 → 146 ms, `format.clj` 131 → 100 ms; output byte-identical on all twelve files, parser suite 141/141.

New rows, base vs this branch vs the JVM, this machine:

| row | base | this branch | JVM |
|---|---:|---:|---:|
| `coll-dispatch` | 280 ms | 136 | 58 |
| `metadata` | 200 | 112 | 73 |
| `parallel-colls` (8 threads) | 536 | 435 | 94 |

Gates: `make recordinline` 49/49 (red on the base tree); gambitgencheck/gambitcheck/gambitkernel/gambitunbound/gambitvars/gambitstatics/gambittwins, portcheck, deadhost, mirrordrift, adaptercheck; `make libconformance` 47 ok / 0 regressed / 3 standing skips; `ci/bench-gate.sh` base vs candidate 41 rows within 1.40x, none slower, 35 faster:

```
mandelbrot     0.59x   nth-access    0.60x   literals      0.63x   dispatch      0.65x
sorted-access  0.65x   sorted-build  0.65x   binary-trees  0.67x   hash-eq       0.69x
string-ops-unh 0.70x   keyed-lookup  0.71x   seqs          0.71x   gc-arrays     0.72x
lazy-threads   0.73x   string-build  0.73x   string-ops    0.73x   typed-rec-unh 0.73x
char-scan-unh  0.74x   char-scan     0.75x   printing      0.77x   apply-rest    0.78x
arrays-unh     0.82x   mono-dispatch 0.84x   vecops        0.85x   cst-format    0.86x
collections    0.87x   arrays        0.88x   byte-arrays   0.88x   loop-recur    0.88x
typed-records  0.89x   transducers   0.90x   compile-forms 0.91x   transients    0.92x
stm            0.93x   executors     0.96x   host-io       0.97x   mathfns       1.00x
mathfns-unh    1.00x   string-scan   1.00x   tak           1.00x   fib           1.01x
startup        1.03x
```

`make test`: the first full run stopped at threadsafety scenario 1 (16 hashing threads, 0/16 done in 120 s); the scenario takes 4.0 s on both the base tree and this one, passes 7/7 alone and after the same preceding gates, so that was the machine, not the change. A full rerun is in flight and its result goes in a comment.


yogthos:
Review. No code changes needed. What I checked beyond reading:

- The wrapper's open-coded predicate is `(record? v 'rtd)`, and I confirmed on Chez 10.4.1 that the two-argument `record?` answers `#t` for an *opaque* type too (only the one-argument form refuses opaque records), so a future `(opaque #t)` layout would not silently break — no host type uses one today.
- The load-order gate is the important half: it makes the values.ss move an invariant rather than a fact about today's tree.
- The ratio gate is measured in one process, both arms best-of-3 with a re-measure band, which is the shape the timing-gate rule allows.
- The `make test` rerun the description promised finished green (`TEST_MAKE_EXIT=0`), and CI's test job on the tip is green on both runners.

Merged after #1033 and #1036: this branch merges perf/meta-field (which merged fix/1031), records-gambit.ss regenerated over both, `make recordinline gambitgencheck` rerun on the merged tree, then `make libconformance` on the final tree.

Issue #1039, opened 2026-09-17
Implement kmet.loader for Jolt


markokocic:
This is related to #912 

I have some WIP code, will create a PR for review once it's polished.

Issue #912, opened 2026-09-09
Provide Classloader lite for Jolt


markokocic:
Provide a possibility to emulate Classloader behavior on Java for dynamic class loading and unloading.

For example, to create a new context that would accept classpath as a parameter. Inside that context, `require` would use that classpath to resolve ns from classes and dependencies. It should also accept a resolve-fn, which  user can provide, which will provide namespaces from the host process, whitelist, blacklist, ....

The benefit is that each new context `require` would be isolated from others, so there would be no clashes between different or incompatible dependencies or classes in different contexts. So `ns1.fn1` in one context can be completely different from `ns1.fn2` in another context, without issues.

It should support creating context, and unloading context.

sci has something similar in sci/init, but it starts the complete interpreter. Java Classloader may be the better approach for base implementation.

markokocic:
hi @yogthos , I drafted a plan to implement classloader like loader concept for Jolt, bb and Clojure. You can find a sketch https://github.com/markokocic/kmet/blob/master/loader.md 

First version would be sci based, that should work everywhere.

For second phase, looks like it would be possible to implement native loader support for Jolt, to be able to write real Jolt code for extensions, not only sci.

Can you please review it, and let me know if that's something you'd consider implementing in Jolt (or accepting a PR)?

yogthos:
Thanks for sketching it out, and the general shape seems right to me. I'd go with the native loader for Jolt from the start here since I don't think it would be a huge amount of extra over over doing a sci version first.

yogthos:
I might actually give it a shot here today and see how it goes.

markokocic:
Well, I need sci anyways for bb, and I can do real testing in kmet right now.

I will be able to test Jolt loader only after I finish porting terminal.

One more observation. kmet.libs.loader is meant to be compat layer on top of Jolt, bb, Clojure and sci, but it still makes sense for Jolt to expose its own jolt.loader protocol hiding scheme internals.

yogthos:
Yeah agreed, your loader can be a common API here. As a side note, SCI should run on the latest version of Jolt as well which should unblock your interpreted approach already.

markokocic:
Cool. That would bring parity with for free. Will give it a try.

But having a proper loader would still be much better.

yogthos:
Yup, I think it'd be generally useful.

PR #1040, opened 2026-09-17
Implement jolt.loader: per-context namespaces, vars and resources

markokocic:
The Loader protocol declared in e4879011 was a surface of throws; this is the implementation behind it. The conformance suite (test/chez/loaderconf-test.clj, run by `make loaderconf`) passes in full — 18 cases, and test/chez/loaderconf-known-failures.txt is empty.

The runtime substrate is one global namespace registry (rt.ss's var-table), so a context is built out of it rather than beside it:

  * a namespace located on a loader's OWN roots is private to that context. Any version already installed under the name is evicted (remove-ns) before the source is evaluated, so the evaluation makes fresh var cells; compiled references are direct links, so the evicted cells stay live for code that already holds them. The root loader hides an owned name — and vars in it — so a second context resolving it does not see the first context's version.
  * a namespace resolved through the DELEGATE stays where the delegate put it, shared by reference; a hit located by a delegate is linked at its home loader too, so the next loader reaching that delegate re-uses the handle instead of re-reading the source.
  * loading a namespace pre-loads its (ns … (:require …)) dependencies through the loader first, and a dependency the loader cannot serve fails the load with an actionable :loader/unreadable. Evaluated source resolves through the loader — policies and the delegate chain included — never through the runtime's global require.
  * require / resolve / ns-resolve / find-var calls in evaluated source are rewritten to context-carrying forms (__require-in and friends) before compilation, so a function loads in the context that DEFINED it, whoever calls it and across a fiber park. A quoted target of resolve/find-var is qualified with the defining namespace while the source is still data: at call time *ns* is the caller's.
  * the evict-evaluate-snapshot window mutates global state under one name, so it is serialized by a per-name claim shared by every loader (waiters block on a promise, so fibers stay parkable); same-thread re-entry raises :loader/circular instead of deadlocking. A failed load leaves no partial namespace, and unload! unmaps the namespaces it installed while they are still the ones installed, drops its ownership, and leaves definitions already resolved live.

The protocol, the generic resolution body, the link table, the in-flight claims, the policy combinators (allow / deny / self-first / pool / isolated), the ambient tier and the host root are the real implementation: `find` never reads, `resolve` touches the link table only, `load` reads once, hits are data and print/compare/travel, and `unload!` is idempotent, non-blocking and reports teardown failures as data.

Host seams for the classloader facade: clojure.java.io/resource's 2-arity resolves through a loader object that answers getResource (java/io.ss consults the library's tagged-table registry through the new tagged-method-lookup), and clojure.lang.RT/baseLoader returns the ambient loader's facade inside with-loader (registered over the host singleton captured at load).

Known limits, recorded in the namespace docstring: private loads mutate the process-global registry, so a context shadowing a host name evicts the host's registration (the runtime's loaded mark survives remove-ns, so a plain require does not restore it — :reload does), and a host-side in-place reload of a context-owned name reuses that context's object and cells; the per-name claim orders writers, not readers; :class requests have no backend yet; jar roots are rejected; and load reads source directly rather than through the AOT cache, because a compiled artifact bakes its cross-namespace links — the host-root path does use the cache, through jolt.host/load-namespace.

Gates: loaderconf 18/18 on the built binary, unit, selfhost, corpus, smoke, buildsmoke, stdlibfasl, manifestcheck, documented, cts, stateimage, aotcachesmoke.

Refs:
- https://github.com/jolt-lang/jolt/issues/912
- https://github.com/jolt-lang/jolt/issues/1039


markokocic:
This PR is now done, at least for phase 1.

It now fully implements https://github.com/kmetia/kmet-agent/tree/master/src/kmet/loader and finally allows Kmet to load extensions on Jolt.

yogthos:
Reviewed against the built binary; 27/27 passed as pushed. Two things I found and fixed on the branch (2ec452aa), plus a merge of main with a remint:

**The source walk could not tell a var call from a local.** `rewrite-context-ops` replaced any `(resolve …)` / `(load …)` / `(require …)` it met in the unexpanded forms, so ordinary extension code broke:

- `(fn [resolve reject] (resolve 42))` — the promise-callback idiom — became `(__resolve-in id 42)`: "Long cannot be cast to Named"
- `(defn total [load] (load 3))` — refused as a host-file `load`
- `(let [require (fn [x] (inc x))] (require 1))` — answered nil

Scope is the compiler's to know, so the rewrite moved there: the analyzer's invoke arm now consults a hook (`jolt.host/*invoke-rewrite*`, `hc-invoke-rewriter` in host-contract.ss) after macro expansion and only when the head is not a local and resolves to a var. `jolt.loader` binds it around a context's evaluation with `context-rewriter`, which builds the same `__x-in` forms you had. Two consequences worth knowing: the `ns` form's own `(clojure.core/require …)` now goes through `__require-in` too (so `:as`/`:refer` land in the defining ns on that path as well; I made it pass `:exclude`/`:rename` through), and a macro-emitted `require` is rewritten at its expansion site, i.e. the compiling context. The hook fires only while the compile namespace is the one being loaded — the binding is dynamic, so a host namespace pulled in transitively at load time (`requiring-resolve` at the top level) compiles inside it, and that code goes through the AOT tee; case 29 pins that it stays the host's (proven red without the ns check).

**A `:var` request naming a class answered a hit whose `:cell` was the class** — `(find (root) {:kind :var :name "clojure.core/String"})` returned `java.lang.String` as the cell. Only a `var?` is a hit now.

Smaller: `with-loader`'s docstring said the binding is inherited by threads; a bare `Thread.` does not convey (only what `binding` conveys to), reworded. `:load? false` is accepted but the source backend always preloads dependencies (compiling needs them); noted in `req-map`'s docstring rather than pretending otherwise. CHANGELOG entry added.

Unrelated to this PR but surfaced by case 29's first draft: jolt pre-registers a runtime `(require '[x :as a])` inside a fn body in the *defining* namespace at compile time (`ce-scan-requires!`), which the JVM does not — filed as jolt-d1ip.

`make test` green locally on the fix; merging main (#1041 landed meanwhile) needed a remint of both seeds. Merging once CI agrees.

PR #1057, opened 2026-09-19
Dispatch (.-name target args*) as a method call, not a field read

### LINKED ISSUE #1056: (.-name target args*) is analyzed as a field read, so the arguments are dropped and the call fails one argument short
## Summary

`(.-name target args*)` is analyzed as a field read and the arguments are
silently discarded. Clojure treats only the **no-argument** form as a field
access; with arguments it is a call of the member named `-name`. That is how a
protocol method whose name genuinely begins with a dash gets invoked.

The dropped arguments do not surface as an arity error at the call site — the
field value is then invoked one argument short, so the error names an anonymous
`fn` and points nowhere near the form that caused it.

## Repro

```clojure
(defprotocol P (-go [this x]))
(deftype T [] P (-go [_ x] (str "called with " x)))

(.-go (T.) 42)    ; JVM 1.12.0: "called with 42"
                  ; jolt:       Wrong number of args (1) passed to: fn

(. (T.) -go 42)   ; "called with 42" on both
(-go (T.) 42)     ; "called with 42" on both
```

Field reads are unaffected:

```clojure
(deftype F [fld] Object)
(.-fld (F. "hello"))   ; "hello" on both
```

## Why it matters

This is what blocks [cognitect aws-api](https://github.com/cognitect-labs/aws-api)
on jolt. Every release from **0.8.603 through 0.8.824** calls

```clojure
(a/<!! (.-invoke-async client op-map))
```

in `cognitect.aws.client.impl`, so `aws/invoke` dies with
`ArityException: Wrong number of args (1) passed to: fn` before any request is
built. Measured across the releases in my `~/.m2`:

| version | `(.-invoke-async …)` on the invoke path | result on jolt v0.8.9 |
|---|---|---|
| 0.8.612, 0.8.692, 0.8.723, 0.8.741 | yes | ArityException |
| 0.8.762, 0.8.824, 0.8.847 | no | loads |

0.8.847 happens to have dropped the pattern, which is the only reason
[jlt-commons/aws-api-jolt](https://github.com/jlt-commons/aws-api-jolt) works
today — it has to pin that exact release. Any library using dash-prefixed
protocol methods hits the same wall.

## Cause

`jolt-core/jolt/analyzer.clj`, `analyze-field`:

```clojure
{:op :host-call
 :method (subs hname 1)        ; ".-field" -> "-field"
 :target (analyze ctx (nth items 1) env)
  :args []}                    ; <- everything after the target is dropped
```

`field-head?` routes every `.-`-prefixed head here regardless of arity, and
`:args` is hardcoded empty.

## Related

#1020 (closed) covers host method **tables** silently ignoring extra trailing
arguments. This is a different code path: the arguments are discarded during
analysis and never reach a method table at all. Same family of symptom,
different mechanism.

## Fix

PR follows: take `(drop 2 items)` as `:args` so the no-argument form stays a
field read and the argument form becomes a method call.

## Verification

jolt v0.8.9, macOS/aarch64. Reference behaviour checked against JVM Clojure
1.12.0.

Verified against `upstream/main` at `4e710c96` on 2026-09-20, current HEAD as of
filing — `analyze-field` there still has `:args []`.

burinc:
Fixes #1056.

## What

`analyze-field` hardcoded `:args []` and discarded everything after the target,
so `(.-name target args*)` was analyzed as a field read of `-name` and the
arguments vanished. The resulting value was then invoked one argument short.

Clojure treats only the no-argument form as a field access. With arguments it is
a call of the member named `-name` — which is how a protocol method whose name
begins with a dash gets invoked.

```clojure
(defprotocol P (-go [this x]))
(deftype T [] P (-go [_ x] (str "called with " x)))

(.-go (T.) 42)    ; before: Wrong number of args (1) passed to: fn
                  ; after:  "called with 42"        (JVM 1.12.0 agrees)
(. (T.) -go 42)   ; correct before and after
(.-fld obj)       ; field read, unchanged
```

The change is `:args (mapv #(analyze ctx % env) (drop 2 items))`. The
no-argument form still yields `:args []`, so field access is untouched.

## The seed

Re-minted per CONTRIBUTING, since the analyzer is compiled into the image.
`make selfhost` reports `rebuild == checked-in seed` on this base.

## Verification

jolt built from this branch on macOS/aarch64, compared against the stock v0.8.9
binary on the same probe.

Gates run on this branch, all passing:

```
unit gate: 1902/1902 passed        infer gate: 48/48 passed
fieldread gate: 11/11 passed       fieldnum gate: 9/9 passed
fieldjoin gate: 12/12 passed       host-derived-props: 68 checks, 0 failures
host-registry: 31 checks, 0 failures
corpus, cts, sci, scifunctional: pass (exit 0)
```

`cts` and `sci` produce **byte-identical** counts with and without this change
(`impl/namespaces.cljc: 120 ok, 7 fail`, `core.cljc: 53 ok, 11 fail`,
`SCI load: 399/424`), so nothing there moves either way.

Real-world check: `com.cognitect.aws/api 0.8.692`, which cannot load on stock
jolt, gets past the `ArityException` at `cognitect/aws/client/api.clj:130` with
this binary.

## One gate that fails, and why it is not this change

`make certify` fails on `java.util.zip` / charset corpus rows (`ZipFile`,
`Deflater`, `ZipInputStream`, `ZipEntry`). I verified this is pre-existing
rather than assuming it: I stashed this change, confirmed the tree was
byte-identical to `upstream/main` (`git diff upstream/main --stat` empty, fix
marker absent), and re-ran `make certify` — the same rows fail.

`certify` evaluates the corpus's `:expected` values against reference JVM
Clojure and never runs the built binary, so an analyzer change cannot affect it.
Worth noting it gates the whole `ci` target, so everything after it is skipped
locally; that is why the list above is the gates run individually.

Verified against `upstream/main` at `4e710c96` on 2026-09-20.


yogthos:
Reviewed; the analyzer change is right and matches what `analyze-dot` already emits for `(. x -go 42)`, and the Chez seed fixpoint holds here too.

CI's `test` job failed on `gambitseedcheck`: the analyzer is a seed source for both hosts and `host/gambit/seed/` is cross-minted from the Chez seed, so it needed re-minting as well. I pushed that mint onto this branch (57652f5d) rather than round-trip it — `make remint` only did the Chez half, which is a docs/tooling gap on our side and is being fixed separately so the next seed-source PR doesn't hit it.

The runtime side of the same family (the field arm still answered a declared slot with the argument dropped, `(.-fld x 1)` → 7, and a dashed miss with arguments was worded as a field miss) plus corpus rows and the changelog entry follow in a stacked PR.

PR #1084, opened 2026-09-21
Task runner: :exec-fn / :cmd CLI tasks, via a vendored babashka/cli

### LINKED ISSUE #1083: Task runner: would you accept `:exec-fn` / `:cmd` CLI tasks, via a vendored babashka/cli?
## The question first

I would like to add babashka's CLI-tasks support to jolt's task runner, and
doing it faithfully needs a vendored copy of `babashka/cli`. That is roughly
2,500 lines of third-party source as a new submodule, so I would rather ask
before opening a large PR than after.

If vendoring is not something you want, the feature is still possible in
smaller or more divergent forms. I have listed those at the bottom with what
each one costs.

## Relationship to #878

This is the follow-up to the limitation #878 named for itself. That issue
shipped `jolt completions zsh|bash|fish` and `jolt completions tasks` in
v0.8.5, completing task **names**, and said plainly that per-task **argument**
completion was out of reach because "jolt has no per-task option metadata today
(babashka gets it from `:exec-fn`/`:cmd` routing through `babashka.cli`)".

This issue is about closing exactly that gap. It is not a proposal to change or
replace what #878 shipped.

## Why it is worth doing

In babashka a task can declare its options, and then gets `--help` and real
shell completion for free. On jolt the same task map runs, but the options are
invisible: `jolt <task> --help` has nothing to print, and completion after the
task name can only offer a filename.

For using jolt as a babashka replacement for scripting, this is the last gap I
hit. Everything else in a task map already works.

## What is missing today

Checked against `main` at `70133c55`:

- `jolt-core/jolt/tasks.clj` supports `:doc`, `:task`, `:depends`, `:requires`,
  `:private`, `:extra-paths`, `:extra-deps` and the `:init`/`:requires`/
  `:enter`/`:leave` hooks. There is no `:exec-fn`, `:cmd`, `:cli`, `:exec-args`
  or `:restrict`. (`grep exec-fn` on that file returns nothing; the three `:cmd`
  hits are an ex-data key on a failed shell command, not a task key.)
- The `:exec-fn` in `jolt-core/jolt/main.clj:590` is the tools.deps alias `-X`
  mechanism, which is a different thing.
- `babashka.cli` is not bundled. `stdlib/babashka/` contains only `tasks.clj`.

## The part that makes this cheap

`babashka/cli` runs on jolt **unmodified**. I checked rather than assumed, with
`0.12.91` on jolt in source mode:

- `parse-opts` with `:coerce` returns the coerced map
- `dispatch` routes a command tree and folds a handler's own
  `:org.babashka/cli` metadata
- `format-opts` renders help text including defaults and enum values
- `:enum` validation rejects with babashka's own wording

So there is no parser to write. The work is wiring an existing library into the
task runner, which in babashka's own implementation is about 515 lines
(`src/babashka/impl/tasks.clj` goes from 473 lines at `v1.12.218` to 941 at
`v1.13.223`).

It also fits jolt's existing shape. `stdlib/jolt/cli.clj` can be
`(import-vars babashka.cli :exclude #{*exit-fn*})`, exactly as
`stdlib/jolt/process.clj` wraps `babashka.process`. Of babashka.cli's 24
public vars, `*exit-fn*` is the only one `import-vars` cannot express, because
it is dynamic and a delegating `defn` wrapper would break `binding`.

## Approach I would propose

1. `vendor/cli` submodule pinned to `v0.12.91`, with `vendor/cli/src` on
   `ldr-install-roots` beside `fs` and `process`. No new gate: `fs` and
   `process` have none, and `grenadinecheck` exists only to reconcile
   grenadine's separate generated-sources tree, which `cli` has no equivalent
   of.
2. `stdlib/jolt/cli.clj` as the public face, with `babashka.cli` still directly
   requirable so a `bb.edn`'s own `(:require [babashka.cli])` resolves.
3. `:exec-fn` / `:cmd` / `:cli` / `:exec-args` routed through
   `babashka.cli/dispatch` from `run-body!`, required lazily so a shell-string
   task does not pay to load an option parser.
4. `--help` short-circuiting before the `:depends` walk, per babashka 1.13.219.
5. The hidden `org.babashka.cli/completions` command special-cased in
   `jolt.main`, matching what babashka does in its own `main.clj`, so a
   generated snippet works verbatim. `jolt completions` stays exactly as
   #878 shipped it.
6. Both seeds re-minted.

Target semantics are babashka 1.13.219 + 1.13.221 + 1.13.223, including #2151.
Anything in babashka's Unreleased section is out of scope.

## Alternatives, if vendoring is unwelcome

- **Copy the source into `stdlib/`.** No submodule, but jolt then
  hand-maintains 2,500 lines that drift from upstream with no mechanical way to
  measure the drift.
- **Resolve `org.babashka/cli` as a Maven dependency on demand.** Cheapest by
  far and adds nothing to the image, but it needs network or a warm `~/.m2` on
  first use, and babashka needs neither. A `bb.edn` that works in babashka and
  fails on an air-gapped jolt is not the parity that makes the feature worth
  having.
- **A jolt-native option parser.** Smaller, and no third-party code in the
  tree, but every difference in coercion, help layout or error wording becomes
  a parity bug that only surfaces in someone else's project.

I have a working branch for the vendored approach, currently two commits in
(the submodule and the shim), and I am happy to reshape it or drop it depending
on what you would actually merge.

---

Verified against `upstream/main` at `70133c55` on 2026-09-22.

burinc:
Closes #1083.

A `bb.edn` task that names an `:exec-fn` — or a `:cmd` tree of them — has its
arguments parsed by `babashka.cli` before anything runs, which is what gives it
coercion, validation, subcommands and `--help`. That is babashka's CLI-task
feature, and this brings it to jolt at **1.13.219 + 1.13.221 + 1.13.223**
semantics.

The PR was opened as direction-seeking, with two commits and the question
"do you want a vendored `babashka/cli` in this tree?" (#1083). The answer is
yes, and the remaining nine tasks of the plan are now here.

## What landed

- **`vendor/cli`** — `babashka/cli` at `v0.12.91`, registered on
  `ldr-install-roots` plus the four other places a vendored source root has to
  appear (`TESTBIN-INPUTS`, three `set-source-roots!*` sites).
- **`stdlib/jolt/cli.clj`** — the public face, following `stdlib/jolt/process.clj`:
  `(import-vars babashka.cli :exclude #{*exit-fn*})`. Of the 24 public vars,
  `*exit-fn*` is the only one `import-vars` cannot re-export, because it is
  dynamic and a delegating `defn` wrapper would break `binding`.
- **`jolt.tasks`** — `cli-node`, `task-node`, `cli-dispatch!` and `run-cli-dep!`,
  keeping babashka's names so `babashka.impl.tasks` and this can be read side by
  side. babashka assembles a program string for SCI; jolt evaluates the task map
  directly, so the same semantics are wired as ordinary function calls.
- **`jolt.completions` + `jolt.main`** — the hidden
  `org.babashka.cli/completions` callback, and zsh/bash/fish snippets that use
  it for CLI tasks only.
- **README + `llms.txt`** — "Tasks that parse their arguments".

## The semantics, measured against babashka

Every row below was run against real `bb` on the same `bb.edn`, not assumed.
`jolt <task> --help` and `bb <task> --help` are byte-identical apart from the
program name, inherited options included.

| behaviour | source |
|---|---|
| `:exec-args` on the task as well as under `:cli`, the bare key winning | 1.13.220 |
| a task's `:cli` `:exec-args` **merge over** the `:tasks` top-level, not replace | 1.13.220 |
| a task's `:cli` spec **adds to** the runner-level spec; `--help` lists the rest under `Inherited options` | 1.13.220 |
| a task with `:exec-fn` runs when another task `:depends` on it | 1.13.220 |
| a dep's options parse for the CLI task that runs, and show in its `--help` | 1.13.220 |
| a CLI task cannot name a `:cmd` task in `:depends` unless it also has a `:task` body | 1.13.220 |
| `:cmd` may be a symbol naming a var holding the tree; its namespace loads on demand | 1.13.220 |
| completion offers inherited options too | 1.13.220 |
| `tasks` and `--help` still describe a task when a dependency's namespace will not load | 1.13.220 |
| a `:depends` `:exec-fn` receives **all** parsed options, runner-level `:cli` defaults included | 1.13.221 |
| with `:restrict`, only the options it declares | 1.13.221 |
| a `:depends` `:restrict` falls back to the runner-level one | 1.13.221 |
| a `:depends` `:exec-fn` receives its own `:exec-args` and spec defaults, command line winning; they do **not** apply to the target | #2103 |
| an `:exec-fn` task in the `:depends` of a task with a `:task` body runs, with its own `:exec-args` and spec defaults | #2151, 1.13.223 |

The three rows the installed `bb` (1.13.220) answers differently are exactly the
1.13.221 / 1.13.223 changes, and jolt takes the newer side of each — checked
against `babashka.impl.tasks` at 1.13.223 rather than against the older binary.

`--help` short-circuits before the `:depends` walk: the body, the dep walk and
the `:enter`/`:leave` pair all go in as thunks the parser calls only once it has
picked a command, so asking what a task accepts runs nothing. Hook placement,
the diamond (a shared CLI dep runs once), `--parallel`, and
`babashka.tasks/run` reaching a CLI task were each checked against `bb` too.

## Two deliberate differences

- **Errors are plain `ex-info`, not `{:babashka/exit 1}`.** In jolt that key
  means "exit with this status, the failure has already reported itself", which
  is right for a failed subprocess and wrong for a mistake in a `bb.edn` — it
  would make an unresolvable `:exec-fn` exit 1 in silence. A bad task map is
  reported the way jolt already reports `unknown command or task`. Exit codes
  match babashka's.
- **No docstring fallback for a task's `:doc`** in `jolt tasks` and completion.
  Both read the project's config files alone on jolt, and deriving that doc
  means loading the handler's namespace — a completing shell must not be what
  discovers your deps do not resolve. A docstring still reaches `--help`, where
  the fn is being resolved anyway.

## Completion costs a plain task nothing

`jolt completions tasks` marks a task that parses with a third `cli` field, so
the snippets call jolt back only for those. What such a task accepts depends on
where the cursor is and cannot be cached as a flat list; every other task still
completes from the cache without starting jolt at all.

## Gates

`clishim` (new, in `CI-GATES`), `taskssmoke` 102, `completionssmoke` 40,
`stdlibfasl` 11, `documented`, `deadhost`, `portcheck`, `depssmoke`,
`scriptsmoke`, plus the full CI matrix. Fixture: `test/chez/tasks/cliproj`.

No gate was added for the submodule itself: `vendor/fs` and `vendor/process` are
plain submodules with none, and `grenadinecheck` exists only to reconcile
grenadine's separate generated-sources tree, which `cli` has no equivalent of.

## Relationship to #878

#878 shipped `jolt completions` in v0.8.5, completing task **names**, and said
per-task **argument** completion was out of reach because "jolt has no per-task
option metadata today". This closes that gap without changing what #878 shipped.


yogthos:
Picked this up and pushed the rest of the plan to this branch — the nine
remaining tasks are done, so the draft is no longer a dependency with no feature
attached. Four commits on top of your three:

**`tasks: dispatch an :exec-fn / :cmd task through babashka.cli`**
The runner itself. `cli-node` / `-task-node` / `-cli-dispatch` / `-run-cli-dep`
ported from `babashka.impl.tasks` at 1.13.223, as ordinary calls rather than
assembled program strings — jolt evaluates the task map directly, so there is no
SCI layer to generate code for. `:exec-fn`, `:cmd` (literal or a symbol naming a
def), `:cli` (map or symbol), `:exec-args` with the task's winning over `:cli`'s,
a spec on the handler var as `:org.babashka/cli` metadata, and a runner-level
`:tasks {:cli ...}`.

Two ordering points, which are the parts worth reviewing:

- `--help` runs nothing. The `:depends` walk, the body and the `:enter`/`:leave`
  pair all go in as thunks and the parser decides whether any is called.
- 1.13.221 / babashka#2151 `:depends`: a CLI dep does not parse; its handler is
  called in its place in the graph with the options the target's parse supplied,
  over its own and the runner-level defaults, narrowed by `:restrict`. Its spec
  merges into the target's, so the option parses there and shows in `--help`.
  `:depends` naming a `:cmd` task is refused.

Errors are plain `ex-info`, not babashka's `{:babashka/exit 1}` — in jolt that
key means "exit with this status, the failure already reported itself", which
would make an unresolvable `:exec-fn` exit 1 in silence.

**`completions: complete a CLI task's own options, without spending a start`**
`jolt.main` now answers the hidden `org.babashka.cli/completions` command. The
snippets do not call it blindly: `completions tasks` gained a third `cli` field,
and zsh/bash/fish call back only for a task carrying it, so a plain task still
completes with no jolt process at all. Two of the smoke rows assert exactly that
by counting spawns.

**`Run the jolt.cli shim test as a gate`**
`test/cli_shim_test.clj` was in the branch but nothing ran it — no target, no
CI-GATES entry. It is `make clishim` now.

**`Document tasks that parse their arguments`** — a README section (there was no
home for the task runner beyond passing references), plus the completion field
and the callback command in `llms.txt`.

Your manifest fix landed while I was working on the same thing; I dropped my
duplicate in the rebase, so `c62362ce` is the one that survives.

One deliberate divergence from babashka, noted in the source: bb falls back to a
handler's docstring for a task with no `:doc` in `bb tasks` and in completion.
Both read config files alone here, and deriving that doc means loading the
handler's namespace — which is the one thing a listing and a completing shell
must not do. The docstring still reaches `--help`.

Gates run locally against the built binary: `taskssmoke` 102/0 (33 new rows over
a new `test/chez/tasks/cliproj` fixture), `completionssmoke` 40/0 (14 new),
`scriptsmoke` 33/0, `stdlibfasl` 11/0, `clishim`, `documented`, `unit`. No seed
re-mint is needed — `emit-image.ss` embeds only the compiler namespaces and
`clojure.core`, not `jolt.tasks`.

Taking it out of draft on that basis; say the word if you would rather keep
iterating on it first.

yogthos:
Answering #1083: **yes** — the vendored `babashka/cli` is wanted, and this is the shape we want it in. Reviewed and taking it.

What I did to the branch:

- **Merged current `main` in** (`568530cd`). It was based on `70133c55`; `main` has since taken #1077, #1078, #1081 and #1085, which move `vendor/grenadine` to 0.1.15 and rework `java/io.ss` and `java/nio-file.ss`. Clean merge, and the merged tree is byte-identical to the rebase I gated with.
- **Retitled and rewrote the description**, which still described the first two commits as direction-seeking groundwork. The remaining nine tasks are in the branch.

What I checked, beyond the gates: every semantic row against real `bb` on the same `bb.edn`, rather than against the changelog. `jolt <task> --help` and `bb <task> --help` come out byte-identical apart from the program name, `Inherited options` included, and so does the completion callback's output. Hook placement, the diamond, `--parallel`, a two-level CLI dependency chain, `:override-builtin` on a CLI task, and a dependency whose handler namespace does not load all agree.

The three rows where the installed `bb` (1.13.220) disagrees are exactly the 1.13.221 and #2151 changes — a `:depends` handler getting all parsed options rather than only its own, `:restrict` keys including the runner level, and a CLI task running under a target that has a `:task` body. This branch takes the newer side of each, which I verified against `babashka.impl.tasks` at 1.13.223 rather than against the older binary. That is the right call.

Local gates on the merged tree: `clishim`, `taskssmoke` 102, `completionssmoke` 40, `stdlibfasl` 11, `documented`, `deadhost`, `portcheck`, `depssmoke` 138, `scriptsmoke` 33.

One follow-up filed rather than fixed here: **`babashka.tasks/exec`** is still missing, so a task body cannot call `(exec 'app.api/build)`. The `cli-node` docstring's "which is what an exec call already accepts" is true of babashka and not yet of jolt; both are on the bead, since the comment becomes true when `exec` lands.

Thanks for opening it as a question first — that was the right way to do it.

